PR #31, opened 2023-04-14
Validate schema using jsonschema

dougiesquire:
This PR gets rid of the janky `metadata.py` module and replaces it with an approach in which schema can be easily configured and validated using `jsonschema`. This should make it easier in the future to read schema for elsewhere, should that be desired.


dougiesquire (commented):
ok

PR #140, opened 2023-11-29
Draft action for automated update of the catalog on Gadi

rbeucher:
Hi @dougiesquire 

This is a draft but I am gonna have to push that to `main` so I can test the workflow...
GitHub does not have the option to run actions from anything else than the main branch.

Is that OK with you?


dougiesquire:
Do what you need to do :)

PR #178, opened 2024-07-22
Extend ACCESS-OM3 filename regex

### LINKED ISSUE #176: ACCESS-OM3 filename patterns are too prescriptive

For ESM1.5 and OM2, cice daily output is concatenated into a monthly files, with filenames containing `-YYYY-MM-daily.nc`

Currently, the OM3 builder accepts filenames of this form (e.g. `om3.cice.h.1900-01-daily.nc`), however the resulting datastore doesn't work. There is a new `file_id` in the datastore for each file, when there should be one for all daily cice data (I believe?)

The regexs used to determine these fileid's needs revisiting:

https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/9381296f1f2c53b5b894ed1e8162f7de70a3073a/src/access_nri_intake/source/utils.py#L195

We could use a different form of the filename for daily cice output

# Additional context
This is needed to add auto post-processing to OM3 configurations (https://github.com/COSIMA/access-om3/issues/182).

To reproduce the error, see https://github.com/anton-seaice/sandbox/blob/main/open_datastore_after_cice_concat.ipynb

dougiesquire:
Closes #176. Decided just to add a quick fix for this. However, this whole process needs to be revisited soon as it clearly isn't scaling to the range of ACCESS outputs.


tests/test_source_utils.py
```diff
@@ -106,49 +106,73 @@
...
-            "GMOM_JRA.mom6.ocean_sfc_1900_01_01",
+            "access-om3.cice.h.1900-01",
+            (
+                "access_om3_cice_h_XXXX_XX",
+                "1900-01",
+                None,
+            ),
+        ),
+        (
+            "access-om3.cice.h.1900-01-daily",
```
anton-seaice, line 133: We should also add "access-om3.cice.h.daily.1900-01" ?

anton-seaice (commented):
I think this won't work if the filename patterns change during the experiment? e.g. half the results use ...1900-XX-XX.nc and the other half use 1900-01-daily.nc ?

dougiesquire:
> I think this won't work if the filename patterns change during the experiment? e.g. half the results use ...1900-XX-XX.nc and the other half use 1900-01-daily.nc ?

I'm not sure what you mean by "won't work" here. They would get distinct `fileids` and so be considered distinct datasets, which I think is the correct behaviour for the way things are set up currently.

dougiesquire:
I think we're now [heading (back) towards using CMIP6-like vocab for the frequency](https://github.com/COSIMA/access-om3/issues/190)

E.g. `access-om3.cice.h.1day.mean.1900-01.nc`

So let's put a pin in the PR until the naming is resolved.

dougiesquire:
Okay, I'll think we've finialised a format - see [here](https://github.com/COSIMA/access-om3/issues/190#issuecomment-2259523255). I'll update this PR.

anton-seaice:
> I'm not sure what you mean by "won't work" here. They would get distinct `fileids` and so be considered distinct datasets, which I think is the correct behaviour for the way things are set up currently.

My concern is that the archive script which turns "access-om3.cice.h.1900-01-01" into "access-om3.cice.h.1900-01-daily" might fail, and then get fixed. Meaning the catalog ends up with two fileids when ideally we want them to be considered the same dataset.

dougiesquire:
> My concern is that the archive script which turns "access-om3.cice.h.1900-01-01" into "access-om3.cice.h.1900-01-daily" might fail, and then get fixed. Meaning the catalog ends up with two fileids when ideally we want them to be considered the same dataset.

Right, this sort of issue can't really be handled with the way dataset grouping is done at the moment. The fix for this currently would be to go back and concatenate the files that were missed. I think that's actually the best solution anyway. But we will be revisiting how the grouping is done soon to make it more robust.

anton-seaice:
I think with `access-om3.cice.1day.mean.1900-01.nc` from https://github.com/COSIMA/access-om3/issues/190#issuecomment-2261753828 we might not need to make a change ?

dougiesquire:
@marc-white I got halfway through this and then got distracted. The issue is that the regex description for ACCESS-OM3 is overly prescriptive and doesn't work with, for example, annual files. Would you be able to take this over? (It may be easier to just close this and start again, given your recent changes)

[This comment](https://github.com/COSIMA/access-om3/issues/190#issuecomment-2272409582) contains a summary of what ACCESS-OM3 output should look like in the not-too-distant future. @anton-seaice can also help answer any questions.

marc-white:
Yep, happy to take over.

minghangli-uni:
Hi @marc-white, I wanted to check in regarding the status of this PR. I am currently having issues with the regex description for ACCESS-OM3, as it doesn’t seem to work with annual files.

marc-white:
@minghangli-uni this got buried under other things, but I'll take a look now.

marc-white:
So is this where the filename formats got to? The original issue is pretty convoluted... https://github.com/COSIMA/access-om3/issues/190#issuecomment-2272409582

marc-white:
@minghangli-uni could you give me some sample filenames that you're trying to work with?

minghangli-uni:
Sure, you can find some in /scratch/tm70/ml0072/access-om3/archive/longerexpt5_rr_mean_snap3-perturb-5873b4db/output000

marc-white:
@minghangli-uni I'm a bit confused, which are the annual files in there you're having issues with? I can only see what appear to be monthly files, with a smattering of others...

minghangli-uni:
I think the problem I am is related to the annual filenames. [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/6be33d8b85c8f43d1ccdf69bb58b8d77e23a927f/src/access_nri_intake/source/builders.py#L427C1-L429C6)

```python
    PATTERNS = [
        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']})$",  # ACCESS-OM3
    ]
```

It now doesnt support something like ['y']?

marc-white:
@minghangli-uni ah yes, I see now. @dougiesquire has already made that update in this pull request, although I can see a couple of other things missing when I compare to your sample directory that I'll need to update.

marc-white:
@minghangli-uni @dougiesquire @anton-seaice is the `GMOM_JRA` filename pattern something we need to keep supporting, or should everything under `ACCESS-OM3` be changed? I saw @dougiesquire had changed all the filename parsing tests away from `GMOM_JRA`, but I'll also need to update some test data within the package if it's really going away.

anton-seaice:
All of our current configs use the names 'access-om3' but I wouldn't rely on that too much. It would be nice to be able to change that name without having to change any code in the catalog builders.

marc-white:
@anton-seaice it's more that the current regex updates for ACCESS-OM3 have broken one of the `GMOM_JRA` patterns, so I need to know if I need to rectify that, or if it's OK to dispose of support for that pattern (plus I'll need to grab some new test data for the catalog creation).

anton-seaice:
Which pattern isn't working ? Its probably fine to dispose of it but also nice if users can make modifications to filenames if they want without breaking things

marc-white:
I managed to work out the issue with the regular expressions, and it seems to now be in a state where it works with both the new `access-om3` and older `GMOM_JRA` patterns.

During this, I ran into an interesting regex conundrum that sucked up a lot of my day, which I wondered if anyone had seen before...

The current `AccessOM3Builder` in main uses an OR-style block to match the date group part of a filename:
```python
PATTERNS = [
        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']})$",  # ACCESS-OM3
    ]
```
(look at the group in parenthesis that references the `'ymds'`, `'ymd'`, etc. keys of `PATTERNS_HELPERS`, with `|` in between each option).

The [re documentation](https://docs.python.org/3.9/library/re.html#regular-expression-syntax) says that `|` searches left to right, taking the first match it finds. So far, so good - we want to see if we can match a Y-M-D-S pattern, if that doesn't work we try to simpler Y-M-D pattern, and so on.

**However**, if I added a Y only pattern to the end of the `|` chain, this would happen:
```ipython
In [42]: b = re.search(AccessOm3Builder.PATTERNS[0], "GMOM_JRA_WD.ww3.hi.1900-01-03-00000.nc")

In [43]: b.group(0)
Out[43]: 'GMOM_JRA_WD.ww3.hi.1900_01_03_00000.nc'

In [44]: b.group(1)
Out[44]: '0000'
```

It's almost like the `|` search suddenly decided to deliberately search for the minimum match possible, and gave me back the last year match it found. Removing the Y pattern gets me back to the result I was expecting:
```ipython
In [32]: b = re.search(AccessOm3Builder.PATTERNS[0], "GMOM_JRA_WD.ww3.hi.1900-01-03-00000.nc")

In [33]: b.group(1)
Out[33]: '1900-01-03-00000'
```

Any thoughts from anyone (but particularly @dougiesquire )?

dougiesquire:
> Any thoughts from anyone (but particularly @dougiesquire )?

I agree that's odd. Unfortunately nothing springs to mind sorry. I'm definitely not a regex expert.

anton-seaice:
Thats pretty odd, I guess you could test it by changing the order of patterns in `AccessOm3Builder.PATTERNS[0]` ?

marc-white:
> Thats pretty odd, I guess you could test it by changing the order of patterns in `AccessOm3Builder.PATTERNS[0]` ?

Yep, tried that. It always seemed to lock on to the "Y" pattern, regardless of where I put it.

src/access_nri_intake/source/builders.py
```diff
@@ -425,7 +425,10 @@ class AccessOm3Builder(BaseBuilder):
     """Intake-ESM datastore builder for ACCESS-OM3 COSIMA datasets"""
 
     PATTERNS = [
-        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']})$",  # ACCESS-OM3
+        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']})(?:$|{PATTERNS_HELPERS['not_multi_digit']}.*)",  # ACCESS-OM3
+        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['y']})(?:$|{PATTERNS_HELPERS['not_multi_digit']}.*)",  # ACCESS-OM3
+        # rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']})(?:$|{PATTERNS_HELPERS['not_multi_digit']}.*)",  # ACCESS-OM3
+        # rf"[^\.]*\.{om3_components}\..*[\.,_]({ymds}|{ymd}|{ym}|{y})(?:$|-{not_multi_digit}.*)" {PATTERNS_HELPERS['y']}
```
minghangli-uni, line 431:
```suggestion
        rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*?({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']}|{PATTERNS_HELPERS['y']})(?:$|{PATTERNS_HELPERS['not_multi_digit']}.*)",  # ACCESS-OM3
```

@marc-white It seems like a non-greedy match can resolve this issue. I've quickly tested some scenarios,

```python
import re
from access_nri_intake.source.builders import AccessOm3Builder

b = re.search(AccessOm3Builder.PATTERNS[0], "access-om3.mom6.3d.uh.1mon.mean.1900.nc")
print(b)
print(b.group(0))
print(b.group(1))

b = re.search(AccessOm3Builder.PATTERNS[0], "access-om3.mom6.3d.uh.1mon.mean.1900-01-01-00000.nc")
print(b)
print(b.group(0))
print(b.group(1))

b = re.search(AccessOm3Builder.PATTERNS[0], "access-om3.mom6.3d.uh.1mon.mean.1900-01.nc")
print(b)
print(b.group(0))
print(b.group(1))

b = re.search(AccessOm3Builder.PATTERNS[0], "GMOM_JRA_WD.ww3.hi.1900-01-03-00000.nc")
print(b)
print(b.group(0))
print(b.group(1))

b = re.search(AccessOm3Builder.PATTERNS[0], "GMOM_JRA_WD.ww3.hi.1900.nc")
print(b)
print(b.group(0))
print(b.group(1))

b = re.search(AccessOm3Builder.PATTERNS[0], "GMOM_JRA_WD.ww3.hi.1900-01.nc")
print(b)
print(b.group(0))
print(b.group(1))
```
gives me outputs,
```
$ python test.py
<re.Match object; span=(0, 39), match='access-om3.mom6.3d.uh.1mon.mean.1900.nc'>
access-om3.mom6.3d.uh.1mon.mean.1900.nc
1900
<re.Match object; span=(0, 51), match='access-om3.mom6.3d.uh.1mon.mean.1900-01-01-00000.>
access-om3.mom6.3d.uh.1mon.mean.1900-01-01-00000.nc
1900-01-01-00000
<re.Match object; span=(0, 42), match='access-om3.mom6.3d.uh.1mon.mean.1900-01.nc'>
access-om3.mom6.3d.uh.1mon.mean.1900-01.nc
1900-01
<re.Match object; span=(0, 38), match='GMOM_JRA_WD.ww3.hi.1900-01-03-00000.nc'>
GMOM_JRA_WD.ww3.hi.1900-01-03-00000.nc
1900-01-03-00000
<re.Match object; span=(0, 26), match='GMOM_JRA_WD.ww3.hi.1900.nc'>
GMOM_JRA_WD.ww3.hi.1900.nc
1900
<re.Match object; span=(0, 29), match='GMOM_JRA_WD.ww3.hi.1900-01.nc'>
GMOM_JRA_WD.ww3.hi.1900-01.nc
1900-01
```

marc-white: @minghangli-uni we'll need to add your test filenames into the proper test suite before we can merge.

marc-white: @minghangli-uni I've implemented your suggestion and added some file name test cases - three of those filenames aren't giving the expected test output for the file frequency (I think). I'll need to come back next week with a fresh mind to confirm I haven't just set up the test wrong.

marc-white:
I found the issue with the new filename patterns for testing - they had a very slightly different frequency pattern that wasn't recognized. I've added that, and everything seems to pass muster now.

charles-turner-1 (approved):
Looks good to me - although admittedly I'm too great with regex's.

minghangli-uni:
Thanks @marc-white. It looks good to me! While I'm not an expert in regex, I am confident about the non-greedy match here. This non-greedy match matches as little as possible after the component name, ensuing it stops when it finds a valid date pattern instead of capturing other things.

minghangli-uni (approved):
Thanks @marc-white. It looks good to me! While I'm not an expert in regex, I am confident about the non-greedy match here. This non-greedy match matches as little as possible after the component name, ensuing it stops when it finds a valid date pattern instead of capturing other things.

PR #180, opened 2024-07-31
Update path to schema

rbeucher:
Reading the doc and came across that link that seems old.


dougiesquire (changes requested):
Thanks for catching this @rbeucher. The link you've changed to isn't right either - see my suggestion. There's also another link that needs replaced so I'll do that in a follow-up commit.

dougiesquire (approved):
LGTM - please feel free to merge

PR #181, opened 2024-08-02
Refactor parse_access_* functions into BaseBuilder class

### LINKED ISSUE #67: Need a better way to get dataset keys
Currently Intake-ESM dataset keys are constructed using a flaky approach of trying to redact time stamp information from filenames to construct a file id that is combined with the frequency to uniquely define a dataset, see e.g. https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/581633e2a67aa20b94091eb744a162c93707144d/src/access_nri_intake/source/utils.py#L71

I'm not sure how to reliably get dataset keys when the data are so messy. Going forward, better solutions might be to require that those generating model output:

- include a `file_id` attribute in files that ids a file as part of a dataset
- at least follow particular format when they have to include time info in the filename (e.g. YYYY-MM-DD or something we can be confident is a date)
- ...

marc-white:
Fixes #67 (or at least improves).

This PR refactors the utils functions `parse_access_filename` and `parse_access_ncfile` to be `@classmethod`s of the `builders.BaseBuilder` class. The idea behind doing this is:

- The functions remain available to external callers as, e.g., `BaseBuilder.parse_access_filename`;
- Having them as `@classmethod` (as opposed to `@staticmethod`) gives access to the underlying class definition, so every `Builder` now has a list of relevant patterns (`PATTERNS`) that is used by default for this builder. This has the effect of causing each `Builder` to have its own unique list of filename regex patterns, preventing cross-contamination when new builders/patterns are added. The patterns list used can be defined by kwarg to the functions if required instead;
- Specifying a custom version of `parse_access_*` for a particular Builder is now easier (although I'm not sure if that's something that will ever need to be done).

Associated tests have also been updated.


dougiesquire (approved):
Thanks for this @marc-white!

I guess the main difference in having these as `@classmethod`s, rather than just having `source.utils.parse_access_ncfile` receive an additional input for the patterns (which are an attribute on each Builder class) is that it allows overriding the `parse_access_ncfile` and/or `parse_access_filename`. I'm not sure whether that will ever be needed, but I guess there's no reason not to add that additional flexibility.

marc-white:
@dougiesquire is there any other process we need to tick off before I merge this back in, or is the accepted review enough?

dougiesquire:
An accepted review and the passing checks is sufficient. Please feel free to merge

PR #189, opened 2024-08-29
185 ship schema with package

### LINKED ISSUE #185: Ship schema with package
Current behaviour is that the schema used to validate the datastore and catalog entries [are downloaded and cached](https://access-nri-intake-catalog.readthedocs.io/en/latest/management/schema.html) the first time `access-nri-intake` is imported - see [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/a07470a7b94ceecc62b98f9f11ac30c11a5c3344/src/access_nri_intake/source/__init__.py#L20-L25) and [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/a07470a7b94ceecc62b98f9f11ac30c11a5c3344/src/access_nri_intake/catalog/__init__.py#L21-L26). This has obvious issues when someone is trying to use `access-nri-intake` for the first time from a computer without network access (e.g. Gadi compute nodes).

To avoid this issue, we probably need to refactor how the schema are handled so that the appropriate schema are shipped with the `access-nri-intake` package.

marc-white:
This PR closes #185 by incorporating the schema descriptions (JSON) into the package. This has been done IAW what I think is current best practice for storing & loading data files within a package (https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources).

Whilst doing this, I found a couple of places where the `data` submodule was doing some ad-hoc local data loads. These have also been updated to use the new pattern.

Open questions:

- Is this how we want to ship schema? How often do the schema need updating, and is it a pain to have to tag a new version every time an update is required?

### DIFF
--- a/src/access_nri_intake/catalog/__init__.py
+++ b/src/access_nri_intake/catalog/__init__.py
@@ -18,8 +18,6 @@ YAML_COLUMN = "yaml"
 NAME_COLUMN = "name"
 TRANSLATOR_GROUPBY_COLUMNS = ["model", "realm", "frequency"]
 
-# SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/experiment-metadata/1-0-2.json"
-# SCHEMA_HASH = "ecb72c1adde3679896ceeca96aa6500d07ea2e05810155ec7a5dc301593c1dc7"
 
 EXP_JSONSCHEMA, CATALOG_JSONSCHEMA = get_jsonschema(
     metadata_file="data/metadata_schema_experiment.json", required=CORE_COLUMNS
--- a/src/access_nri_intake/catalog/__init__.py
+++ b/src/access_nri_intake/catalog/__init__.py
@@ -18,11 +18,11 @@ YAML_COLUMN = "yaml"
 NAME_COLUMN = "name"
 TRANSLATOR_GROUPBY_COLUMNS = ["model", "realm", "frequency"]
 
-SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/experiment-metadata/1-0-2.json"
-SCHEMA_HASH = "ecb72c1adde3679896ceeca96aa6500d07ea2e05810155ec7a5dc301593c1dc7"
+# SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/experiment-metadata/1-0-2.json"
+# SCHEMA_HASH = "ecb72c1adde3679896ceeca96aa6500d07ea2e05810155ec7a5dc301593c1dc7"
 
 EXP_JSONSCHEMA, CATALOG_JSONSCHEMA = get_jsonschema(
-    url=SCHEMA_URL, known_hash=SCHEMA_HASH, required=CORE_COLUMNS
+    metadata_file="data/metadata_schema_experiment.json", required=CORE_COLUMNS
 )
 
 COLUMNS_WITH_ITERABLES = [
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -216,6 +216,7 @@ def build():
 
     _here = os.path.abspath(os.path.dirname(__file__))
     if update:
+        # FIXME use correct package data handling
         with open(os.path.join(_here, "data", "catalog.yaml"), "w") as fobj:
             yaml.dump(yaml_dict, fobj)
 
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -4,6 +4,7 @@
 """ Command line interfaces for access-nri-intake """
 
 import argparse
+import importlib
 import logging
 import os
 import re
@@ -214,10 +215,12 @@ def build():
     # Save the catalog
     cm.save()
 
-    _here = os.path.abspath(os.path.dirname(__file__))
     if update:
-        # FIXME use correct package data handling
-        with open(os.path.join(_here, "data", "catalog.yaml"), "w") as fobj:
+        with (
+            importlib.resources.files("access_nri_intake")
+            .joinpath("data/catalog.yaml")
+            .open(mode="w") as fobj
+        ):
             yaml.dump(yaml_dict, fobj)
 
 
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -4,7 +4,6 @@
 """ Command line interfaces for access-nri-intake """
 
 import argparse
-import importlib
 import logging
 import os
 import re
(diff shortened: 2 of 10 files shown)


src/access_nri_intake/data/metadata_schema_experiment.json
```diff
@@ -0,0 +1,168 @@
+{
```
marc-white, line 1: I removed the `$id` value from here (and from the other schema) because it just pointed to the GitHub raw location of the file - should this be reinstated?

src/access_nri_intake/cli.py
```diff
@@ -214,9 +215,12 @@ def _get_project(path):
...
     cm.save()
 
-    _here = os.path.abspath(os.path.dirname(__file__))
     if update:
-        with open(os.path.join(_here, "data", "catalog.yaml"), "w") as fobj:
+        with (
+            importlib.resources.files("access_nri_intake")
+            .joinpath("data/catalog.yaml")
+            .open(mode="w") as fobj
+        ):
```
marc-white, line 223: This change does not get touched by the test suite, and I haven't been able to manually test it... A look at it would be appreciated.

src/access_nri_intake/catalog/__init__.py
```diff
@@ -18,11 +18,11 @@
 NAME_COLUMN = "name"
 TRANSLATOR_GROUPBY_COLUMNS = ["model", "realm", "frequency"]
 
-SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/experiment-metadata/1-0-2.json"
-SCHEMA_HASH = "ecb72c1adde3679896ceeca96aa6500d07ea2e05810155ec7a5dc301593c1dc7"
+# SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/experiment-metadata/1-0-2.json"
+# SCHEMA_HASH = "ecb72c1adde3679896ceeca96aa6500d07ea2e05810155ec7a5dc301593c1dc7"
```
aidanheerdegen, line 22: Best to just delete these lines if no longer required.

src/access_nri_intake/cli.py
aidanheerdegen: @jo-basevi do you have any suggestions for how this might be tested?

src/access_nri_intake/data/__init__.py
```diff
@@ -1,9 +1,11 @@
...
+import importlib
 
 import intake
 
-_here = os.path.abspath(os.path.dirname(__file__))
-data = intake.open_catalog(os.path.join(_here, "catalog.yaml")).access_nri
+data = intake.open_catalog(
+    importlib.resources.files("access_nri_intake").joinpath("data/catalog.yaml")
+    # os.path.join(_here, "catalog.yaml")
+).access_nri
```
aidanheerdegen, line 11: As this code is duplicated would it be a good idea to make it a function that can be used in the two locations? Would that help with testing?

src/access_nri_intake/data/metadata_schema_experiment.json
aidanheerdegen: Honestly not sure. Any comment @CodeGat or @jo-basevi?

src/access_nri_intake/source/__init__.py
```diff
@@ -17,9 +17,9 @@
 PATH_COLUMN = "path"
 VARIABLE_COLUMN = "variable"
 
-SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/file-metadata/1-0-1.json"
-SCHEMA_HASH = "8f2f069fa06d81ff086b91daa6503f75615aa90385ab61ee2d1a7956dc96f9a6"
+# SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/e9055da95093ec2faa555c090fc5af17923d1566/au.org.access-nri/model/output/file-metadata/1-0-1.json"
+# SCHEMA_HASH = "8f2f069fa06d81ff086b91daa6503f75615aa90385ab61ee2d1a7956dc96f9a6"
```
aidanheerdegen, line 21: As before, best to delete.

aidanheerdegen (commented):
Some comments and pinged Jo and Tommy for comments.

src/access_nri_intake/data/metadata_schema_experiment.json
jo-basevi: I think it was added initially as setting `$id` to an URL as an identifier was mentioned as best practice in json-schema docs (https://json-schema.org/understanding-json-schema/basics#declaring-a-unique-identifier). I think `$id` is only strictly necessary when referencing a schema from another (https://www.learnjsonschema.com/2020-12/core/id/), which isn't the case here. But maybe if future iterations of the schema use that functionality, it might be worth keeping it in for now? 

 Other thoughts to keep it are that it is consistent with other schemas in the [ACCESS-NRI/schema](https://github.com/ACCESS-NRI/schema) repository, and it self-documents an URL that retrieves the schema?

CodeGat: I think keep it for the above reasons.

src/access_nri_intake/cli.py
jo-basevi: Sorry I don't have any good ideas on testing. Could have a helper function in the upper-level `utils.py` file return `importlib.resources.files("access_nri_intake").joinpath("data/catalog.yaml")` so that can be reused and that can be tested? Might be worth having global vars for  the "access_nri_intake" package name and the filenames.

I didn't know about `importlib.resources` so that was great to learn about!

marc-white: My main concern is the last line of this block, the `mode="w"`. I'm not sure how the system will like trying to write back to something in its own `data` directory, although I suspect as this is a standard-ish file open, it's probably OK? I might just have to rig up a way to manually test that the write-back is respected, and then test on the catalog location finding only.

src/access_nri_intake/data/__init__.py
marc-white: It could, although I'm going to have to do some beard-stroking as I implement it. This block loads the existing catalog; the block referenced above (cli.py, L213-223) updates the stored catalog. I'll just need to make sure whatever helper I come up with has actually been loaded in both of those situations.

marc-white: I've updated to a helper function with a related test, but I need to think further about how to test:

- The catalog update in `cli.py`, and
- The `intake` catalog generation here. This step looks for `/g/data` access, so I'm going to have to come up with a way to mock that.

aidanheerdegen:
Let me know when you want another review @marc-white

src/access_nri_intake/data/__init__.py
marc-white: I'll resolve this particular comment thread as the importlib call has been factored out into its own testable function, but I'll repeat the above comments about testing during the next review pass.

aidanheerdegen (approved):
LGTM. Just some commented code to delete.

src/access_nri_intake/data/__init__.py
```diff
@@ -1,9 +1,14 @@
 # Copyright 2023 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
 # SPDX-License-Identifier: Apache-2.0
 
-import os
-
+# import importlib
+# import importlib.resources
```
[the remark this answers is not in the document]
marc-white: Done.

CodeGat (approved):

Issue #186, opened 2024-08-26
Return type for parse_access_ncfile


marc-white:
# Is your feature request related to a problem? Please describe.

The return out of `BaseBuilder.parse_access_ncfile` is currently a long tuple of assorted file information. This makes it difficult to remember which item is which during debugging & testing.

# Describe the feature you'd like

We should consider if `parse_access_ncfile` should return a `dict` instead. This will have the advantage of making the return more human-legible. (It will also simplify the `parser` functions of the various builders, which effectively all rebuild the tuple return into a `dict` anyway.)

# Describe alternatives you've considered

Nil.

# Additional context

I'm about to add another issue (#187) to suggest an expansion of the testing for builders, which will be made easier by this feature.

charles-turner-1:
The changes in #212 should cover this - I'd initially thought that we'd want to have separate flags in the catalogue for data and coordinate variables, so I factored it out as a way of keeping track of changes. It turned out to be unnecessary to do that, but I agree the long tuple is difficult to keep track of & creates a lot of unnecessary context.

I used a dataclass, not a dict, however - let me move a lot of the repetitive stuff inside the class. Given pydantic is used pretty heavily upstream in intake-esm, I wonder whether we might want to use a pydantic dataclass? It might be handy to enforce the extra validation?

Issue #117, opened 2023-09-22
[BUG] Can't load 'area_t' from ACCESS-OM2


anton-seaice:
# Describe the bug

The access-nri-intake catalog doesn't open the tracer area field from ACCESS-OM2 results. If I try and load the 'area_t' variable from '01deg_jra55v140_iaf_cycle4_jra55v150_extension' I get a failed to load error.

# To Reproduce

Run:
`catalog['01deg_jra55v140_iaf_cycle4_jra55v150_extension'].search(variable='area_t').to_dask()`

Actual Result:
`Failed to load dataset with key='ocean_2d_area_t.fx'`

Expected Result:
Dataset with x and y dimensions, and area_t variable.

# Additional context

I suspect this is because there are multiple files with the area_t variable, but they are all the same (i.e. it's a field for information/analysis, not explicitly model output).

i.e., these files cannot be concatenated (with default settings) by xarray because they are all the same
`['/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1000/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1001/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1002/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1003/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1004/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1005/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1006/ocean/ocean-2d-area_t.nc',
 '/g/data/ik11/outputs/access-om2-01/01deg_jra55v140_iaf_cycle4_jra55v150_extension/output1007/ocean/ocean-2d-area_t.nc',
[...]

Issue #201, opened 2024-09-27
Making coordinate variables searchable


charles-turner-1:
# Is your feature request related to a problem? Please describe.

Currently, the ACCESS-NRI Intake Catalog doesn't allow for searching of coordinate variables: for example, searching for `st_edges_ocean` will return 0 datasets. This can make searching for coordinate variables difficult, with 2 main pain points:
1. If the coordinate variable is know to be stored in a netCDF file with specific data variables:
    - The user needs to know that these data & coordinate variables are found in the same files, and then search for the data variable in order to access the coordinate variables.
    - In some instances, the user can directly access the coordinate variables by searching the data variables. In others, they need to perform something like 
```python
        VARNAME = ...
        COORD = ...
        fname = cat.search(variable=VARNAME).df.loc[0,'file']
        ds = xr.open_dataset(fname)
        coord = ds[COORD]
```
Although this doesn't require the user to (semi) manually work out what file to open, it's still messy as it requires passing round file names.

2. In other instances, coordinate variables are stored completely separately. For example, `ocean_grid.nc` files only contain coordinate variables, and so *cannot* be found using the catalogue. The only way to currently access these files is to search the catalogue to get a handle on the directory structure - and then construct a file path and load it: eg:
```python

        fname = cat.search(variable=VARNAME).df.loc[0,'file']
        dirname = Path(fname).parent
        grid_file, = [file for file in os.listdir(dirname) if file.endswith('_grid.nc')] 
        ds = xr.open_dataset(grid_file)
        coord = ds[COORD]
```
This requires the user to start poking round in directory structures to try to work out where to load their data - which is the problem intake is trying to solve.

This has caused some pain points migrating COSIMA recipes from [cosima_cookbook => intake](https://github.com/COSIMA/cosima-recipes/issues/313). 

I also think this might be the same issue as discussed in #63? @aidanheerdegen - seem to be some concerns about coordinates being listed as variables when they shouldn't be there?

# Describe the feature you'd like

Searchable coordinates: in the same way that the catalog currently lets you perform searches over variables, it would be useful to be able to do the same on coordinates: 
```python
var = cat.search(variable=VARN).to_dask()
coord = cat.search(coord=COORD).to_dask()
```

Doing this is subject to a couple of constraints:
1. The catalog needs to know that coordinates & data variables aren't the same & need to be treated differently - `xr.combine_by_coords` will fail if passed a coordinate variable.
[...]

Issue #112, opened 2023-08-18
Encoding grid information


aidanheerdegen:
# Is your feature request related to a problem? Please describe.

The resolution of a dataset is an important piece of information. It can be critical when searching for data to know the resolution, as the representation of physical processes is typically dictated by the resolution of a model. The information on the resolution of a dataset is encoded in the underlying grid coordinates. 

Also when comparing datasets knowing what other datasets use the same grid is very useful information, as it allows a comparisons to be made without any time consuming, and often technically demanding, regridding.

# Describe the feature you'd like

I want everything, but would be content with a system that extracts grid information when the netCDF files are opened and inspected during the cataloguing process, and saves that grid information in a form that can be queried.

One suggestion is to save grid information into a complementary catalog or tool that can be queried independently of the main ACCESS-NRI Intake Catalog, somewhat similar to the variable suggester tool (#26)

There are number of increasingly convoluted thought bubbles about how to uniquely identify grids in [this issue](https://github.com/COSIMA/cosima-cookbook/issues/191), but the gist is this:

Assuming there are `md5` checksums attributes for all coordinate variables (ideally a independent post-processing step) something like this pseudo-code:

```python
for variable in indexed_variables:
    for coord in variable.coords:
        grid = []
        if coord.attr['md5'] not in coordinates:
            coordinates[coord.attr['md5']] = coord
        grid.append[coord.attr['md5']]
    if grid not in grids:
        grid_id = grids.add(grid)
    grid_id = grids[grid]
    dataset.add(grid_id)
```

Where `coordinates` and `grids` would be serialised to a catalog that could be queried. The `dataset` catalog is already serialised, but the `grid_id`'s found in `dataset `would be added to `dataset` metadata.

So queries could be done to retrieve which datasets contained a given `grid`, and it should be possible to provide a function or the logic required to say if two datasets share a common grid.

An issue for the MOM data is that masked data (which is most of it) also has masked coordinates. In an ideal world a post-processing step would fix that, but it hasn't been done in the past so will need to be supported. 

In the linked COSIMA Cookbook issue [I suggested](https://github.com/COSIMA/cosima-cookbook/issues/191#issuecomment-685332529) we could augment the grid information with a `metadata.yaml` file that could give grids useful human-readable names, but also define relationships between grids, perhaps defining the grids with missing data to be equivalent to unmasked grids. 

Apologies if I've used the wrong terminology above, e.g. datasets, and so made this needlessly confusing.

[...]

PR #212, opened 2024-10-03
Making coordinate variables searchable 

charles-turner-1:
## Functionality changes in this branch:
- `source.builders.BaseBuilder.parse_access_ncfile` indexes all variables (`for var in ds.data_vars` => `for var in ds.variables`), rather than just data variables, & tests have been updated to reflect additional data obtained.
    - NOTE: This is dependent on changes to [intake-esm](https://github.com/charles-turner-1/intake-esm/tree/issue_660) proposed by @dougiesquire in [isssue 660](https://github.com/intake/intake-esm/issues/660).
    - Coordinate variables are accessible through the variable flag in the access-nri intake catalogue, rather than a separate coord flag, as suggested in #201 by @anton-seaice.
        - When searching for coordinate variables only, only coordinate variables are returned. No concatenation of assets is performed - only the first asset in the dataset is. This avoids crashing resulting from `xr.combine_by_coords`.
        - When searching for coordinate & data_variables, the data & coordinate variables searched, as well as coordinate variables that the searched data variables depend on.
        - Other searches will return all data & coordinate variables, as before. 
   
## Refactoring
- Updated `source.builders.BaseBuilder` to use a dataclass (`source.utils._AccessNCFileInfo`) to pass around information about variables, rather than an outputs tuple.
- Added a `_DataVarInfo` dataclass to hold attribute information from data variables, rather than appending them all into separate lists passed to the outputs tuple.

## Minor changes
- Add a bunch of type checking (type annotations, added mypy to pre-commit, etc)
- Replaced clauses of the pattern
    ```python
    try:
        x = x[key]
    except:
       pass
    ``` 
   with
    ```
    x = x.get(key,x)
    ```
- A couple of minor docstring updates

### intake-esm
The code changes in this branch should lead to catalogues being built with coordinate variables included in their variable flag: eg
```python 

from access_nri_intake.source.builders import AccessOm2Builder

builder = AccessOm2Builder(
    path="/g/data/ik11/outputs/access-om2-01/01deg_jra55v13_ryf9091"
)

builder.build()

CATALOG_PATH = ...

builder.save(
    directory='CATALOG_PATH',
    name="ACCESS-OM2_01deg_jra55v13_ryf9091", 
    description="...",
)
```
will build a catalog which coordinate variables are searchable from. However, unless you're working with a version of intake-esm which also has coordinate variables searchable (alterations noted in Functionality changes), this will present two issues:

1. Files comprising *only* coordinate variables, eg. `ocean_grid.nc` files, will not be indexed.
2. Searchs purely for a coordinate variable will fail when calling `.to_dask()` will fail on coordinate indexing if a multi file dataset is obtained from the search, or return an empty dataset if a single file is returned. 

Issue 1. appears to be [ESMDataSource._open_dataset](https://github.com/intake/intake-esm/blob/6ba67e1df72ebbc80b6cd5f78fd73ddb2cee564a/intake_esm/source.py#L219) thinking the .nc files are empty unless they contain coordinate variables & ignoring them.
Issue 2. is the result of the [requested_variables](https://github.com/intake/intake-esm/blob/6ba67e1df72ebbc80b6cd5f78fd73ddb2cee564a/intake_esm/source.py#L84) used in intake-esm. 

Neither of these is testable or fixable in access-nri-intake-catalog, as these issues are handled by intake-esm.

### PR size
This PR contains a couple of code changes, as well as some other stuff (typing/pre-commit, etc). I'm happy to split this apart into separate pull requests if that makes things more tractable to review.


charles-turner-1:
I've been through and linked a bunch of issues that I think the changes here resolve - about to verify.

marc-white (commented):
Correct me if I'm wrong, but this PR doesn't actually do anything to read the coordinate variables itself (compared to the read in the `main` version of the code currently does) - it's simply been prepped to handle the coordinates once `intake-esm` gets updated?

src/access_nri_intake/source/builders.py
```diff
@@ -289,51 +314,31 @@ def parse_access_ncfile(cls, file, time_dim="time"):
...
         ) as ds:
-            variable_list = []
-            variable_long_name_list = []
-            variable_standard_name_list = []
-            variable_cell_methods_list = []
-            variable_units_list = []
-            for var in ds.data_vars:
+            dvars = _DataVarInfo()
+
+            for var in ds.variables:
```
charles-turner-1, line 319: New behaviour - scans all variables in dataset, including coordinate variables

src/access_nri_intake/source/builders.py
```diff
@@ -289,51 +314,31 @@ def parse_access_ncfile(cls, file, time_dim="time"):
             decode_times=False,
             decode_coords=False,
         ) as ds:
-            variable_list = []
-            variable_long_name_list = []
-            variable_standard_name_list = []
-            variable_cell_methods_list = []
-            variable_units_list = []
-            for var in ds.data_vars:
```
charles-turner-1, line 297: Previous behaviour - scans all data_vars in dataset

charles-turner-1:
Not precisely - it does additionally scan .nc files for coordinates (see above), but crucially also requires the esm_datastore object to do the same. If we have the same behaviour in the esm_datastore (defined in intake-esm) there this PR won't affect the variables we read.

charles-turner-1:
I've checked, and the changes in this PR don't resolve #117 or #112.

In both instances, the issues specifically result from attempted concatenation of **grid** variables, rather than coordinate variables. Grid variables are stored as data variables in the netCDF files that hold them, but are don't depend on time & so the time variable is dropped before concatenation, which is the source of the 'no dimension to concatenate along' errors. 

I've changed this PR & the concomitant PR on intake-esm to draft status whilst I work out how closely related these issues are & whether we can neatly tie together all the changes we need.

src/access_nri_intake/source/utils.py
```diff
@@ -4,16 +4,96 @@
...
+    standard_name_list: list[str] = field(default_factory=list)
+    cell_methods_list: list[str] = field(default_factory=list)
+    units_list: list[str] = field(default_factory=list)
+
+    def append_attrs(self, var: str, attrs: dict) -> None:
+        """
+        Append attributes to the DataVarInfo object, if the attribute has a
+        'long_name' key.
+
+        TODO: Why do we need a long name key? seems important
```
marc-white, line 71: Is the code using the presence of 'long_name' as a proxy for "this is a real variable, not just additional info"?

charles-turner-1:
I've split this into a series of stacked PR's: 
#221: Adds typing, mypy type check to pre-commit. Merge into main, addresses #13.
#222: Refactors `BaseBuilder.parse_access_ncfile` to return a dataclass, rather than a tuple. Closes #186 and the PR merges into the branch introduced by #221 
#212: (This PR) - changes `data_vars` => `variables` and updates the relevant tests to ensure we are catching all the correct coordinate variables. PR merges into the branch introduced by #222.

marc-white (approved):
Approved as ready, although are we in a position to merge before Intake-ESM is updated?

charles-turner-1:
I'll go back and triple check before I merge, but it should be the case that the catalogue will build exactly as before if we don't update intake-esm.

charles-turner-1:
The concomitant changes have been merged into intake-esm ([link](https://github.com/intake/intake-esm/pull/681#event-14968541423)), so we should be close to ready to go with making this the default behaviour.

rbeucher:
Great 👍

PR #211, opened 2024-10-03
Add support for BARPA py18 data store

rbeucher:
This is my first attempt at adding an NCI ESM Datastore to the catalog.
I have started playing with the translators which, I think, could be refactored.

Issue #200, opened 2024-09-25
[BUG] Missing metadata.yaml files for existing CMIP5 & CMIP6 catalogues


marc-white:
# Describe the bug

During investigations on #197 , it was found that the directory containing the `metadata.yaml` files for the existing CMIP5 and CMIP6 catalogues (`cmip5_al33`, `cmip5_rr3`, `cmip6_fs38`, `cmip6_oi10`) has gone missing. The directory referred to in the `access_nri_intake_catalog` config (`/g/data/tm70/intake`) no longer exists.

# To Reproduce

See `/g/data/tm70`.

# Additional context

The metadata can be recovered from the existing catalog via, e.g., 

```python

import intake
cat = intake.cat.access_nri
cat["cmpi6_fs38"].metadata

```

marc-white:
Storing existing metadata here to avoid future loss:

#### `cmip5_al33`

```python
{'contact': 'NCI',
 'created': None,
 'description': 'Replicated CMIP5-era datasets catalogued by NCI',
 'email': 'help@nci.org.au',
 'experiment_uuid': '658c95cc-c299-450c-82a1-b2b2308f7c6e',
 'keywords': ['cmip'],
 'license': None,
 'long_description': 'All CMIP5-era replicated data contained under the project al33.  All file versions present are in the listing. Maintained By: NCI Contact: help@nci.org.au References: https://pcmdi.llnl.gov/mips/cmip5/',
 'model': ['CMIP5'],
 'name': 'cmip5_al33',
 'nominal_resolution': [None],
 'notes': 'null',
 'parent_experiment': None,
 'reference': None,
 'related_experiments': [None],
 'url': 'https://geonetwork.nci.org.au/geonetwork/srv/eng/catalog.search#/metadata/f9489_5106_5649_5038',
 'version': None,
 'catalog_dir': ''}
```

#### `cmip5_rr3`

```python
{'contact': 'NCI',
 'created': None,
 'description': 'Australian CMIP5-era datasets catalogued by NCI',
 'email': 'help@nci.org.au',
 'experiment_uuid': '473d0c44-ab66-458c-b32e-1e1774175853',
 'keywords': ['cmip'],
 'license': None,
 'long_description': 'All CMIP5-era Australian published data contained under the project rr3.  All file versions present are in the listing. Maintained By: NCI Contact: help@nci.org.au References: https://pcmdi.llnl.gov/mips/cmip5/',
[...]

PR #219, opened 2024-10-14
199 data request 20km regional projections for cordex cmip6 queensland future climate science

### LINKED ISSUE #199: [DATA REQUEST] 20km regional projections for CORDEX-CMIP6 from the Queensland Future Climate Science Program
# Description of the data product

This dataset includes projections at 20km, formatted to meet the CORDEX-CMIP6 data standards. The 20km projections were derived from the 10km projections.

# Location of the data product on Gadi

Catalog available in /g/data/ig45
# Checklist

Add a "x" between the brackets to all that apply
- [X] This data product is stable (unlikely to change substantially or move)
- [X] This data product is of use to the broader community
- [X] This data product is documented: [link](https://geonetwork.nci.org.au/geonetwork/srv/eng/catalog.search#/metadata/f7465_8388_5100_7022)
- [X] This data product is licensed under Creative Commons Attribution 4.0 International
- [X] Those who want to access this data can be added to the project that houses it

charles-turner-1:
+ Added Cordex CMIP6 Translators 

+ Beginning refactoring @rbeucher noted would be useful in #211. There's a little more work to do here so I've left this as a draft PR for now.

@marc-white You added `config/experiments/cmip6_ig45.yaml` in this branch - I've separately added a `config/metadata_sources/cordex_ig45.yaml` which is doing largely the same thing but seemed to follow the other source conventions more closely. Are you able to confirm whether there's a convention I'm unaware of?  

Closes #199


marc-white (commented):
@charles-turner-1 there isn't a convention for naming the metadata sources that we're storing in the repository. This is a very recent change that was made to avoid repeating issues like #200. However, they're only reference copies, and the catalog itself should reference the `metadata.yaml` stored somewhere on `/g/data`. Therefore, you can really do what you like with this, so long as we stay internally consistent.

config/experiments/cmip6_ig45/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+  This dataset includes projections at 20km, formatted to meet the CORDEX-CMIP6 data standards. 
+  The 20km projections were derived from the 10km projections.
+model:
+- CMIP6
+frequency:
+- 
+variable:
+- 
+nominal_resolution:
+-
```
marc-white, line 14: You can just leave these as blank-blank, rather than a blank array.

config/experiments/cmip6_ig45/metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+- 
+version: 
+contact: NCI
+email: help@nci.org.au
+reference:
+license:
+url: https://geonetwork.nci.org.au/geonetwork/srv/eng/catalog.search#/metadata/f7465_8388_5100_7022
+parent_experiment:
+related_experiments:
+-
```
marc-white, line 23: As above

config/cmip6.yaml
```diff
@@ -11,3 +11,7 @@ sources:
   - metadata_yaml: /g/data/xp65/admin/access-nri-intake-catalog/config/metadata_sources/cmip6-oi10/metadata.yaml
     path:
       - /g/data/oi10/catalog/v2/esm/catalog.json
+
+  - metadata_yaml: /g/data/xp65/admin/intake/metadata/cmip6_ig45/metadata.yaml
+    path:
+    - /g/data/ig45/catalog/v2/esm/catalog.json
```
marc-white, line 17: If this is the path where the 'live' `metadata.yaml` is being stored, then I'd recommend the reference copy's name/path within the repo match, i.e. be `cmip6_ig45`.

config/cordex.yaml
```diff
@@ -0,0 +1,9 @@
+builder: null
+
+translator: CordexTranslator
+
+sources:
+
+  - metadata_yaml: /g/data/xp65/admin/access-nri-intake-catalog/config/metadata_sources/cordex-ig45/metadata.yaml
```
marc-white, line 7: I don't think this is a path where we want to be storing live `YAML`... see discussion on #200

marc-white (commented):
As it currently stands, it looks like you're going to load the same Intake-ESM catalogue into `access-nri-intake-catalog` twice, using two different translators and `metadata.yaml` files. I presume you wanted to use the `cordex` ones in the final merge?

charles-turner-1:
Yup - the changes to `cmip6.yaml` seem to go with the draft `metadata.yaml` you added. The cordex experiments needed a different translator, so I've separated them out - just hadn't cleaned up that file - everything should be sensible now.

I'm not 100% sure I understand the discussion around YAML paths fully - either that or it appears we need to update a bunch of paths in `config/*.yaml`?

marc-white:
@charles-turner-1 There's a few things about YAML paths:

- Typically, the `metadata.yaml` lives with the data. However, there are some "historical" experiments for which we don't have write access to their data area, so we have to store the `metadata.yaml` within our own `xp65` project. Where in `xp65` was debated, given there is also an `access-nri-intake-catalog` checkout in there somewhere we don't want to fiddle with/be dependent on in production.
- There were some experiments that we were already holding the metadata for in `xp65`, but it accidentally got deleted. That's why we're holding all the metadata for the live catalog in a single place (`/g/data/xp65/admin/intake/metadata`) to try and stop that happening again.
- **Separately**, @rbeucher asked that we start holding a reference copy of new experiment `metadata.yaml`s within the `access-nri-intake-catalog` repository. This **isn't** for production use; it's just for reference in case the above problem happens again (and I guess might be handy for testing too, now that I think about it).

config/cordex.yaml
charles-turner-1, line 1: @marc-white Can you confirm this is where we are wanting to store our metadata files? Just wanna double check I've understood correctly.

config/cmip6.yaml
```diff
@@ -10,4 +10,4 @@ sources:
     
   - metadata_yaml: /g/data/xp65/admin/access-nri-intake-catalog/config/metadata_sources/cmip6-oi10/metadata.yaml
     path:
-      - /g/data/oi10/catalog/v2/esm/catalog.json
+      - /g/data/oi10/catalog/v2/esm/catalog.json
```
rbeucher, line 13: Add empty line at the end of the file

config/cordex.yaml
rbeucher: Yes that is correct.

config/metadata_sources/cordex-ig45/metadata.yaml
```diff
@@ -0,0 +1,126 @@
...
+contact: NCI
+email: help@nci.org.au
+reference:
+license:
+url: https://geonetwork.nci.org.au/geonetwork/srv/eng/catalog.search#/metadata/f7465_8388_5100_7022
+parent_experiment:
+related_experiments:
+notes: 
+keywords:
+- cmip
```
rbeucher, line 126: Add empty line at the end of the file

tests/data/esm_datastore/cordex-ig45.json
```diff
@@ -0,0 +1,70 @@
...
+            "column_name": "variable_id"
+        },
+        {
+            "column_name": "version"
+        },
+        {
+            "column_name": "time_range"
+        }
+    ]
+}
```
rbeucher, line 70: Add empty line at the end of the file

rbeucher (commented):
I think we need to move out of draft test to run the tests.

charles-turner-1:
@rbeucher my bad, moved it out of draft now.

rbeucher (approved):

PR #221, opened 2024-10-15
13-typing

charles-turner-1:
This branch separates out all the type hints & related changes from #212.

Includes:
- Updated gitignore to include VS Code artifacts.
- Added mypy to pre-commit
- Updated a few docstrings to include missing arguments & additional info about raised errors

@marc-white are you able to review for me? I'm in the process of peeling apart the changes from #212 and will stack the refactor & then change of `data_vars` => `variables` on this.


.pre-commit-config.yaml
```diff
@@ -10,3 +10,10 @@ repos:
     hooks:
     - id: black
       language_version: python3
+# Mypy
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v1.11.2'  
+    hooks:
+    - id: mypy 
+      name: mypy 
+      additional_dependencies: [types-PyYAML==6.0.12.20240808]
```
marc-white, line 19: What is the effect of this?

charles-turner-1: Adds a mypy check to pre-commit - where type annotations are added, this will reject changes if they're incorrect.

charles-turner-1: N.B if there aren't any type annotations added, this won't affect anything.

src/access_nri_intake/source/builders.py
```diff
@@ -378,9 +397,12 @@ def __init__(self, path):
...
 
     @classmethod
-    def parser(cls, file):
+    def parser(cls, file) -> dict:
         try:
-            match_groups = re.match(r".*/output\d+/([^/]*)/.*\.nc", file).groups()
+            # mypy gets upset as match can return None. I assume this is why we
+            # have try/except block in the first place? If so, we might be able
+            # to make this more explicit?
+            match_groups = re.match(r".*/output\d+/([^/]*)/.*\.nc", file).groups()  # type: ignore
```
marc-white, line 405: It looks to me like that idea is that we're just palming off any error as 'invalid asset', which is why the `except` clause catches basically everything.

charles-turner-1: Yeah - I'm generally in favour of making try/except blocks as tight as possible, but if we can potentially have issues from the parser too lets leave this as is for now. Just pushed an update to this block.

marc-white (approved):
No issues here, although you might want to flag that too-wide `except` catch for future action with a TODO/FIXME.

charles-turner-1:
Good call - will add it on a dependent branch though, can't remember 100% but I think changing it here will trigger a review being needed again.

PR #222, opened 2024-10-15
Refactor BaseBuilder.parse_access_ncfile

### LINKED ISSUE #186: Return type for parse_access_ncfile

charles-turner-1:
Closes #186, and builds on #221 .

### Changes:
Added two new dataclasses to `src/access_nri_intake/source/utils.py`:
- `_VarInfo`: Dataclass that holds information related to a single variable in a given netCDF file: lists of variable names, cell methods, units ,etc, which default to empty lists. It also contains a method to append to these lists (`append_attrs`), and a method to dump its contents to a dictionary (`to_var_info_dict`), which is used in the constructor of `_AccessNCFileInfo`.
- `_AccessNCFileInfo`: Dataclass to hold information from the netCDF file, containing information about all fields in the netCDF, as well as filename, path, etc. It also has a `to_dict` method which dumps the dataclass to a dictionary, following patterns previously used in parser methods of the different builers.


src/access_nri_intake/source/builders.py
```diff
@@ -286,19 +293,18 @@ def parse_access_ncfile(cls, fname: str, time_dim: str = "time") -> tuple:
 
         Returns
         -------
-        tuple
-            A tuple containing the information parsed from the file
+        output_nc_info: AccessNCFileInfo
```
marc-white, line 296: The class is technically `_AccessNCFileInfo` (note underscore) - will not having that break things if we, e.g., auto-generate documentation?

src/access_nri_intake/source/builders.py
```diff
@@ -590,42 +527,18 @@ def parser(cls, file):
...
-                "variable_cell_methods": variable_cell_methods_list,
-                "variable_units": variable_units_list,
-                "filename": filename,
-                "file_id": file_id,
-            }
-
-            return info
+            ncinfo_dict["file_id"] = re.sub(exp_id, "", ncinfo_dict["file_id"]).strip(
+                "_"
+            )
```
marc-white, line 537: Not critical, but I think this line-breaking/indentation looks a bit better:

```suggestion
            ncinfo_dict["file_id"] = re.sub(
                exp_id, "", ncinfo_dict["file_id"]
                ).strip("_")
```

src/access_nri_intake/source/utils.py
```diff
@@ -15,6 +17,80 @@ class EmptyFileError(Exception):
...
+class _AccessNCFileInfo:
+    """
+    Holds information about a NetCDF file that is used to create an intake-esm
+    catalog entry.
+
+    ______
+    Notes:
+    Use of both path and filename seems redundant, but constructing filename from
+    the path using a __post_init__ method makes testing more difficult. On balance,
+    more explicit tests are probably more important than the slight redundancy.
```
marc-white, line 30: I don't mind a bit of redundancy here - it's a negligible additional thing we need to track that makes life a bit easier (and then only at catalog build time, which is already a ~1hr HPC job).

src/access_nri_intake/source/builders.py
charles-turner-1: Good catch

src/access_nri_intake/source/utils.py
```diff
@@ -77,6 +153,8 @@ def get_timeinfo(
         The dataset to parse the time info from
     filename_frequency: str
         Frequency as determined from the filename
+    filename_frequency: str
+        Frequency as determined from the filename
```
marc-white, line 157: I think there's been a merge error here...

src/access_nri_intake/source/builders.py
charles-turner-1: I agree - the formatter results here are definitely difficult to read. How does this look to you?
```suggestion
ncinfo_dict["file_id"] = re.sub(
                exp_id,
                "",
                ncinfo_dict["file_id"],
            ).strip("_")
```

marc-white (approved):
Looks all set

PR #228, opened 2024-10-23
226 editable install

### LINKED ISSUE #226: [BUG] `tests/test_data.py::test_get_catalog_fp` fails unless package is installed as editable.
# Describe the bug

In `tests/test_data.py`, we have the following test:
```python
def test_get_catalog_fp():
    _oneup = os.path.abspath(os.path.dirname("../"))
    assert str(get_catalog_fp()) == str(
        os.path.join(
            _oneup, "access-nri-intake-catalog/src/access_nri_intake/data/catalog.yaml"
        )
    )
```
This test passes when run using `pytest tests/test_data.py::test_get_catalog_fp` in an editable installation (`pip install -e .`), but fails if `access_nri_intake_catalog` (`pip install .`) is installed into site-packages using a regular installation.

# To Reproduce
## Test passing
```bash
$ git clone https://github.com/ACCESS-NRI/access-nri-intake-catalog/ && cd access-nri-intake-catalog 
$ conda create -f ci/environment.yml
$ pip install -e . # ***CRUCIAL DIFFERENCE***
$ pip list | rg intake
access_nri_intake         0.1.4+13.ge90f843.dirty /Users/u1166368/catalog/access-nri-intake-catalog
intake                    0.7.0
...
$ pytest tests/test_data.py
Test session starts (platform: darwin, Python 3.12.6, pytest 8.3.3, pytest-sugar 1.0.0)
rootdir: /Users/u1166368/catalog/access-nri-intake-catalog
configfile: pyproject.toml
plugins: cov-5.0.0, sugar-1.0.0
collected 1 item                                                                                                                                              

 tests/test_data.py ✓                                                                                                                           100%

Results (0.46s):
       1 passed
```
## Test failing
```bash
$ git clone https://github.com/ACCESS-NRI/access-nri-intake-catalog/ && cd access-nri-intake-catalog 
$ conda create -f ci/environment.yml
$ pip install .  # ***CRUCIAL DIFFERENCE***
$ pip list | rg intake
access_nri_intake         0.1.4+13.ge90f843.dirty
intake                    0.7.0
...
$ pytest tests/test_data.py
Test session starts (platform: darwin, Python 3.12.6, pytest 8.3.3, pytest-sugar 1.0.0)
rootdir: /Users/u1166368/catalog/access-nri-intake-catalog
configfile: pyproject.toml
plugins: cov-5.0.0, sugar-1.0.0
collected 1 item                                                                                                                                              


――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_get_catalog_fp ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

    def test_get_catalog_fp():
        _oneup = os.path.abspath(os.path.dirname("../"))
>       assert str(get_catalog_fp()) == str(
            os.path.join(
                _oneup, "access-nri-intake-catalog/src/access_nri_intake/data/catalog.yaml"
            )
        )
E       AssertionError: assert '/Users/u1166.../catalog.yaml' == '/Users/u1166.../catalog.yaml'
E         
E         - /Users/u1166368/catalog/access-nri-intake-catalog/src/access_nri_intake/data/catalog.yaml
E         ?                 ^^^^^                    ^^^ ^^ ^  ^
E         + /Users/u1166368/miniforge3/envs/access-nri-intake/lib/python3.12/site-packages/access_nri_intake/data/catalog.yaml
E         ?                 ^^^^^ + +++++++                  ^^^^^^^ ^ ^^^^^  ^^^^^^ +++++

tests/test_data.py:11: AssertionError

 tests/test_data.py ⨯                                                                                                                           100% ██████████/Users/u1166368/miniforge3/envs/access-nri-intake/lib/python3.12/site-packages/coverage/control.py:894: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")


---------- coverage: platform darwin, python 3.12.6-final-0 ----------
Coverage XML written to file coverage.xml

=================================================================== short test summary info ===================================================================
FAILED tests/test_data.py::test_get_catalog_fp - AssertionError: assert '/Users/u1166.../catalog.yaml' == '/Users/u1166.../catalog.yaml'

Results (0.52s):
       1 failed
         - tests/test_data.py:9 test_get_catalog_fp
```
# Additional context

This bug is arising because we've (in essence - there are some subtleties here) hardcoded paths to respect an editable install - when using a [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) the package structure changes when pip installed.

- This test isn't failing on CI - I've checked and this is because our actions are installing access_nri_intake_catalogue as an editable package and running tests there. This is probably the more concerning issue to be honest - this means that our tests aren't representative of the package  our users are installing. See [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/development_mode.html): 
> Attention

> Editable installs are not a perfect replacement for regular installs in a test environment. When in doubt, please test your projects as installed via a regular wheel. There are tools in the Python ecosystem, like [tox](https://pypi.org/project/tox) or [nox](https://pypi.org/project/nox), that can help you with that (when used with appropriate configuration).

@marc-white It looks to me like you wrote this test - do you know if you had to change the CI to an editable install to get this test to pass on CI or whether it was already set to editable?

I also think bugs like this might put us in favour of using a flat over src layout as an organisational standard - this sort of issue is less common with a flat layout as setuptools wont restructure the package when it builds it (@rbeucher @aidanheerdegen maybe one for the python standards discussion).

charles-turner-1:
- Fixed issue where `tests/test_data::test_get_catalog_fp()` would only work when performed using an editable installation.

Closes #226

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -2,14 +2,20 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+from pathlib import Path
 
+import access_nri_intake
 from access_nri_intake.utils import get_catalog_fp
 
 
 def test_get_catalog_fp():
-    _oneup = os.path.abspath(os.path.dirname("../"))
-    assert str(get_catalog_fp()) == str(
-        os.path.join(
-            _oneup, "access-nri-intake-catalog/src/access_nri_intake/data/catalog.yaml"
-        )
-    )
+    """
+    Check that we're getting the correct path to the catalog.yaml file. We need
+    to ensure that this works both in editable & non-editable installs.
+    """
+    INSTALL_DIR = Path(access_nri_intake.__file__).parent
+    expected_path = os.path.join(INSTALL_DIR, "data/catalog.yaml")
+
+    catalog_fullpath = get_catalog_fp()
+
+    assert str(catalog_fullpath) == expected_path
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install source
         shell: bash -l {0}
-        run: python -m pip install -e .
+        run: python -m pip install .
 
       - name: List installed packages
         shell: bash -l {0}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,11 @@ jobs:
 
       - name: Run tests
         shell: bash -l {0}
-        run: python -m pytest -s .
+        run: coverage run -m --source=access_nri_intake pytest
+      
+      - name: Generate coverage report
+        shell: bash -l {0}
+        run: coverage xml
 
       - name: Upload code coverage
         uses: codecov/codecov-action@v4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,13 +58,10 @@ versionfile_build = "access_nri_intake/_version.py"
 tag_prefix = "v"
 parentdir_prefix = "access-nri-intake-"
 
-[tool.pytest.ini_options]
-addopts = "--cov=./src --cov-report=xml"
-
 [tool.coverage.run]
 omit = [
-    "src/access_nri_intake/_version.py",
-    "src/access_nri_intake/data/__init__.py",
+    "*/_version.py",
+    "*/data/__init__.py",
 ]
 
 [tool.ruff]
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -14,6 +14,7 @@ from access_nri_intake.cli import (
     MetadataCheckError,
     _check_build_args,
     build,
+    metadata_template,
     metadata_validate,
 )
 
@@ -163,3 +164,7 @@ def test_metadata_validate_no_file(mockargs):
     with pytest.raises(FileNotFoundError) as excinfo:
         metadata_validate()
     assert "No such file(s)" in str(excinfo.value)
+
+
+def test_metadata_template():
+    metadata_template()


charles-turner-1:
`codecov/project` seems to be failing due to the fact that there are no tests associated with commit `cf0b3dc11f1bf7a380775fff1d57b40cc5e4b1c6`. I'm not sure that there's any way around that.

tests/test_data.py
```diff
@@ -2,14 +2,20 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+from pathlib import Path
 
+import access_nri_intake
 from access_nri_intake.utils import get_catalog_fp
```
marc-white, line 8: This breaks if you don't have the package installed at all; are we concerned about this? I think there's meant to be some settings that can be added to `pyproject.toml` to fix this, but I couldn't get them to work last time I tried...

charles-turner-1: Hmmmm, good question. I'm guessing it probably doesn't matter - surely the following line will fail anyway if the package isn't installed at all.

In fact, I'm not sure that any sort of testing works if a package isn't installed - it's necessary that we import the functionality we're testing from somewhere, right?

marc-white: I vaguely remember having been able to do some testing on another package way back in the day without bothering with an install - I *think* there's some setting that sticks the local codebase into the top of `PYTHONPATH` when you invoke `pytest`. That being said, I've not been able to remind myself of what it was yet, so it's entirely possible I'm gibbering rubbish.

charles-turner-1: Maybe [this](https://stackoverflow.com/a/54392877)? Pretty sure doing it this way would work too though?

charles-turner-1:
It seems that getting a proper coverage report with a regular (non-editable) installation can cause issues with codecov (see [here](https://community.codecov.com/t/codecov-report-zero-coverage-for-the-code-but-full-coverage-for-the-tests/3392)).

Marked this as draft until I get this fixed.

charles-turner-1:
Issue isn't with codecov but with pytest-cov itself: [see ongoing issue](https://github.com/pytest-dev/pytest-cov/issues/388).

tests/test_data.py
marc-white: Ah yes, I had seen that one before. However, I thought there was a switch/setting you could throw at the top configuration level that had the same effect as `python -m`... might have a poke around this morning.

marc-white: I'm looking at a Stromlo project I've been working on ([AnniesLasso](https://github.com/ANU-RSAA/AnniesLasso)), and I can `pytest` that without having an install of the code. I can't dins anything in the setup of that repo that allows this, so I'm guessing the only reason I can do it is because of the repo structure (the tests are contains inside the main module code).

marc-white (approved):
Good to go. I've managed to run `pytest` against both an editable and non-editable install, although I still can't make `python -m pytest` work against uninstalled code.

charles-turner-1:
> Good to go. I've managed to run `pytest` against both an editable and non-editable install, although I still can't make `python -m pytest` work against uninstalled code.

Are you able to try ```PYTHONPATH=$(realpath ./src) python -m pytest``` from the project root directory for me and tell me what happens? It breaks `tests/test_cli::test_entrypoint` for me (exit code 32512) but otherwise works fine. TBH I wouldn't expect an entrypoint to work properly for uninstalled code so I'm not too concerned about that.

I'm reasonably confident you're right with the tests in AnniesLasso working due to the project structure on uninstalled code - I think it's relying on the default behaviour of the Python interpreter to append `os.getcwd()` to `sys.path` on startup. In fact, I don't think having the tests in the project itself is necessarily what does the trick - again, I think it's a flat layout thing.

I'll leave this one for now - if you get the same results as me with the `PYTHONPATH` variable set, I'll merge it.

marc-white (commented):
> Are you able to try PYTHONPATH=$(realpath ./src) python -m pytest from the project root directory for me and tell me what happens? It breaks tests/test_cli::test_entrypoint for me (exit code 32512) but otherwise works fine. TBH I wouldn't expect an entrypoint to work properly for uninstalled code so I'm not too concerned about that.

Confirmed I get the same result as you do.

PR #231, opened 2024-10-28
Metadata validation provides more complete report

### LINKED ISSUE #216: Make `metadata-validate` provide a full report on first pass

In checking new `metadata.yaml` files, the `metadata-validate` helper script returns a `ValidationError` on the first issue that it encounters. This means that, if you have a metadata file with multiple errors, you only get notified of the first one. You fix it, run `metadata-validate` again, and then repeat the process _ad nauseum_ until all errors are dealt with.

I'd like `metadata-validate` to record all the errors it finds, and then return with a report on all of the errors that it found in one shot.

Nil.

# Additional context

I've not looked yet, but I presume there's an internal function of metadata validation that `metadata-validate` is wrapping. It's an open question in my mind if this update should be applied only to the `metadata-validate` script, or to the underlying function as well.

marc-white:
Closes #216 .

This PR modifies the behaviour of the function `utils.validate_against.schema` so that a report of all metadata schema errors is provided, rather than just the first issue found.

The function still raises `jsonschema.ValidationError`, but the error message is now formatted as follows:

```
01 | <field name or (missing) > : <reason for failure>
02 | <field name or (missing) > : <reason for failure>
...
```

For example, the error message for a metadata file that has a bad value for `name`, and no value at all for `experiment_uuid`, would be:

```
01 | name : None is not of type 'string'
02 | (missing) : 'experiment_uuid' is a required property
```

This behaviour then flows through to the `metadata-validate` command-line script. The returns of this script have also been refined to:

- More clearly show a success message;
- Print out the new report in the case of a validation error;
- To provide stack trace output in the event of any other form of error/exception.

Tests for the new error handling have also been included. These tests are exhaustive, but not complete (I didn't think it was necessary to provide every single possible permutation).


src/access_nri_intake/utils.py
```diff
@@ -102,7 +102,21 @@ def validate_against_schema(instance: dict, schema: dict) -> None:
         Validator, type_checker=type_checker
     )
 
-    TupleAllowingValidator(schema).validate(instance)
+    issues = list(TupleAllowingValidator(schema).iter_errors(instance))
+
+    if len(issues) > 0:
+        issue_str = ""
+        for i, issue in enumerate(issues):
```
charles-turner-1, line 109: passing enumerate a `start=1` argument would remove the need for the `i+1` on lines 112, 115

marc-white: Done

src/access_nri_intake/utils.py
```diff
@@ -102,7 +102,21 @@ def validate_against_schema(instance: dict, schema: dict) -> None:
...
+            try:
+                issue_str += (
+                    f"\n{i+1:02d} | {issue.absolute_path[0]} : { issue.message }"
+                )
+            except IndexError:  # Must be a missing keyword, not a bad type/value
+                issue_str += f"\n{i+1:02d} | (missing) : { issue.message }"
+        # raise jsonschema.ValidationError(textwrap.fill(textwrap.dedent(issue_str)))
```
charles-turner-1, line 114: I think this commented out line is redundant? Looks like WIP to me

marc-white: Yep, removed

charles-turner-1:
I've checked out to this branch & run the tests & all the additional tests seem to be failing on my machine - I'd just like to do a tad more prodding to work out why before I approve the changes. Bit strange given they're working on CI...

tests/test_utils.py
```diff
@@ -68,3 +71,145 @@ def test_validate_against_schema(instance):
...
+        (
+            "bad_metadata/metadata-bad-email.yaml",
+            1,
+            [
+                "email",
+            ],
+        ),
+    ],
+)
+def test_bad_metadata_validation_raises(test_data, instance, no_errs, bad_kw):
```
charles-turner-1, line 201: I think it would be better to rewrite this as 
```python
    fpath = str(test_data / Path(instance))

    with pytest.raises(jsonschema.ValidationError) as excinfo:
        _ = load_metadata_yaml(fpath, EXP_JSONSCHEMA)

    err_msg = excinfo.value.message

    assert (
        f"{no_errs:02d}" in err_msg and f"{no_errs+1:02d}" not in err_msg
    ), "Can't see the right number of errors!"
    for kw in bad_kw:
        assert f"| {kw}" in err_msg, f"Can't see expected specific error for {kw}"
```
& let pytest handle the exception

charles-turner-1: Noticed this whilst investigating local machine test failures - it was just an environment conflict.

Other than this, looks ready to merge.

marc-white: You mean let pytest just spit out the unexpected exception if it doesn't send back `jsonschema.ValidationError` for whatever reason?

charles-turner-1: More or less - it looks to me like lines 212-215 you're manually handling all exceptions other than jsonschema.Validation error (sorry, highlighted the wrong lines in the original comment) in order to raise that we got an unexpected exception? 

It feels cleaner to me to let pytest handle that for us.

marc-white: OK that's done.

marc-white:
@charles-turner-1 what was the environment issue you ran into?

charles-turner-1:
@marc-white just a pytest extension (either watcher or sugar, not sure which - I rebuilt the environment without either) capturing the exception and assuming it was wrapped with exception.value, rather than a bare exception - caused errors to get turned into TypeErrors.

charles-turner-1 (approved):

Issue #235, opened 2024-10-30
Configuration: CLI utility & TUI config tool


charles-turner-1:
# Is your feature request related to a problem? Please describe.

Following the discussion in #191, it looks like it would be useful to add a CLI/TUI tool to create & pre-populate a configuration that allows the user to select different catalog versions, as well as (potentially) some additional configuration.

# Describe the feature you'd like

A console script that creates a pre-populated configuration file, as well as the ability to enter a TUI which walks the user through configuration if desired.

# Describe alternatives you've considered

See discussion in #191.

# Additional context

- I plan on using [textual](https://textual.textualize.io/) to build the TUI.
- I'm going to keep this completely separate for #191 for the time being, & just build a generic tool skeleton - we can then fit it to the configuation.

rbeucher:
Not sure about the TUI. I think we should keep it simple for now.

Issue #191, opened 2024-08-30
Default to a "latest" version?


dougiesquire:
The ACCESS-NRI Intake catalog is versioned in the same way as the `access-nri-intake` package (see `/g/data/xp65/public/apps/access-nri-intake-catalog`).

The default version is equal to the version of `access-nri-intake` being used, though the version can be specified when the catalog is loaded, e.g.

```python
import intake

cat = intake.cat.access_nri
cat.metadata["version"]

cat_another_version = intake.cat.access_nri(version="v0.1.0")
cat_another_version.metadata["version"]
```

This default behaviour ensures that the catalog version and the version of `access-nri-intake` being used are compatible. However, it also means that users with old versions of `access-nri-intake` installed won't use the most recent version of the catalog by default.

We should consider instead defaulting to a "latest" version of the catalog that symlinks to the most recent released version. To manage compatibility issues, we could have "latest" versions for each major release (e.g. `latest_v1`) and ensure we increment the major version when changes are introduced to `access-nri-intake` that break compatibility with old catalogs.

marc-white:
This sounds like an eminently sensible idea. I can think of a few ways to do this in practice:
1. We could store a catalog version (separate to the `access-nri-intake-catalog` version) within the package. The package would then load the newest catalog with that catalog version by default (and catalogs would be created with the current catalog version). It could also auto-reject any attempt to load a catalog with an incompatible version number.
2. We set up a convention where we only increment the major version number of `access-nri-intake-catalog` when we make a change that will invalidate previous catalogs, and select which catalogs to open based on that.
3. The catalog itself could store which version of `access-nri-intake-catalog` it is known to be compatible with/was generated with. We could then have each version of `access-nri-intake-catalog` know which back-versions of itself that generate compatible catalogs.

There are a few questions around any implementation:

- How many catalog versions currently exist? (Gadi is down, else I would just check for myself.) And of those, are there any that are currently incompatible with the up-to-date `access-nri-intake-catalog`?
- What is the purpose of opening an out-of-date catalog? Will we ever hit the point where there's something removed from the catalog that people will want to have access to (and if that's the case, why would we pull it from the catalog)?

dougiesquire:
> We could store a catalog version (separate to the access-nri-intake-catalog version) within the package.

[...]

PR #237, opened 2024-10-30
235: Configuration file

charles-turner-1:
+ Added a catalog-configure console script
+ Tests for behaviour
    - Copies a config.yaml file from `$ROOT/data/config.yaml` to `$HOME/.catalog/config.yaml`.
    - If `$HOME/.catalog/config.yaml`, require -f/--force flag to copy.
    - To specify version, we use the -v/--version flag.
    - To specify interactive mode, we use -i/--interactive flag (currently disabled/ raises SystemExit).


charles-turner-1:
Closes #235.

@rbeucher I haven't added an interactive mode but I think a (very) basic TUI which steps the user through the config file, shows them current values/defaults & lets them update it if they wish would like would improve this feature.

I also haven't added any validation of the version number as I don't know what sort of versioning we're planning to use for this.

marc-white:
I'm not clear on what this configuration file is actually specifying. Is it trying to set a default value for the catalog version, i.e., set an intrinsic default for 

```python
cat = intake.cat.access_nri(version=NNN)
```

charles-turner-1:
Yup, I think so - I haven't added any of the functionality to address the catalog versioning, but the aim would be to have the default catalog version set (in `data/config.yaml`). 

If the user has specified a catalog version they're after, using eg. `$ catalog-configure -v  X.X.X`, then they'll have a `~/.catalog/config` file - which we can then get the package to read preferentially over `data/config.yaml` to load that specific catalog version.

I think that was where the the discussion was leading in #191? I may have completely misunderstood...

marc-white:
I think there's been a crossed wire somewhere here. A user doesn't _necessarily_ need a separate config file to be able to access different catalog versions - simply passing the `version` kwarg to `intake.cat.access_nri` should do it.

**However**, this rule is broken if the catalog versions are different in structure (`columns_with_iterables`), location of the catalog CSV file (`path`), storage requirements (`metadata.storage`), etc. Then, a different `catalog.yaml` will need to be hooked into `access_nri_intake_catalog.data.data` to get the `intake.cat.access_nri` load working correctly.

I can imagine these `catalog.yaml` files living in the same `/g/data/xp65` location at the catalog CSVs do, and the package can be told that's where they are. The question is then, what to do if, for whatever reason, you can't access `xp65`?

- As @rbeucher has suggested back in #191 , a user might want to be able to provide their own specification for the catalog, for example if they've set up their own mini-ACCESS_NRI catalog in their own storage for some reason. In that case, they'd need a system to be able to point `access_nri.data.data` at the right place. That being said, I'm not entirely sure why a user would want to create their own version of `access_nri_intake_catalog` - conceivably, if they know what they want out of their own data, they could just access it directly...
- If neither of these options are found, we could fail out with some sort of Catalog Not Found error.

If we do want to provide the user with a template `catalog.yaml` to be dumped into their home area, it should be a template, or a reference to some location where a `catalog.yaml` exists.

charles-turner-1:
Yup, it looks an awful lot like I've misunderstood something crucial here. I'm gonna go back through and make sure I understand exactly what's going on at the point of intake.cat.access_nri attribute access.

I've marked this as draft for now, rather than closing it - might be something of use here.

PR #241, opened 2024-11-04
240: Add support for Python 3.12

### LINKED ISSUE #240: Python 3.12 (+3.13?) support

As of the lastest intake-esm release ([2024.2.6](https://github.com/intake/intake-esm/releases/tag/v2024.2.6)), intake-esm no longer supports Python 3.9.

Currently, this package supports 3.9,3.10,3.11, leaving us with the option to use either 3.10 or 3.11. 

Add 3.12 support (and perhaps 3.13).

# Additional context

- Nothing in the test suite breaks when running tests on Python3.12 (on my machine). I haven't tested on CI against 3.12 (Ubuntu 22.04 I think versus local MacOS).
- (N.B) Deprecating 3.9 would add support for typing syntactic sugar, eg. 
```python
def foo(x : float, y : float | None) -> float:
    """ Return x + y if y is a float, else x """
    if y is not None:
        return x + y
    return x
```
Previously, this would have had to have been written as 
```python
from typing import optional
def foo(x : float, y : Optional[float]) -> float:
    """ Return x + y if y is a float, else x """
    if y is not None:
        return x + y
    return x
```
This is potentially confusing *as y is not an optional argument*, it is optional as to whether it has a value. With that said, Python 3.9 is not deprecated until October 2025, so we should really keep supporting it.

charles-turner-1:
Closes #240 

- Added `ci/environment-3.12.yml`.
- Updated `.github/workflows/ci.yml` to include 3.12 run.
- Updated tox.ini.

`tox.ini` is currently broken for Python 3.12. I'm 99% sure this is a tox config issue, not an issue with the package, but leaving this as draft until the tox configuration is fixed.


rbeucher:
Can we add 3.13?

charles-turner-1:
3.13 Works fine on my machine, it's failing to build the environment for 3.13 on CI though.

Looking into it now.

rbeucher:
you are pinning to 3.12 in the environment file

charles-turner-1:
Whoops - must have committed before saving.

charles-turner-1:
3.13 looks good to go.

rbeucher (approved):

rbeucher:
Feel free to merge

PR #247, opened 2024-11-13
Handle case of creating new-style catalog when old-style catalog already exists

### LINKED ISSUE #246: [BUG] handle case where there is an existing catalog that doesn't conform to new catalog format
# Describe the bug

Of course I realized something just *after* the PR #243 was merged... (following on from #191)

The catalog builder is going to have issues the first time we run it, because there is an existing catalog in the live directory, but it doesn't match the new `catalog.yaml` structure. Specifically, there isn't a `min` or `max` version in the old catalog, so `cli::build` will throw a `KeyError` when it tries to read them.

# To Reproduce

Will only occur if a live build process is triggered (not recommended right now).

# Additional context

We could, of course, solve this outside code by simply deleting the existing `catalog.yaml` before we do our first new catalog build. However, this would leave users inexplicably unable to access `intake.cat.access_nri` for a few hours while that happens.

My solution will be to effectively treat this situation as if there is no existing `catalog.yaml`, but without having to cripple the system while the new catalog builds.

marc-white:
Fixes #246 .

This PR handles the case where a new-style catalog (vYYYY-MM-DD) is built, but an old-style catalog (vX.X.X) is present in the catalog build directory.

The code will now treat this the same as a no-catalog-exists case (new `catalog.yaml`, set min and max version based on the catalog directories present), with the exception that the storage flags from the old-style catalog will be merged in.

### DIFF
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -347,7 +347,9 @@ def _combine_storage_flags(a: str, b: str) -> str:
     """
     aflags = re.findall(STORAGE_FLAG_PATTERN, a)
     bflags = re.findall(STORAGE_FLAG_PATTERN, b)
-    return "+".join(set(aflags + bflags))
+    # Sorting the return aids in testing & comparison,
+    # plus makes it more human-readable/human-searchable
+    return "+".join(sorted(list(set(aflags + bflags))))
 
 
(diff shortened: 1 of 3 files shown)


charles-turner-1 (approved):
Looks good

Issue #223, opened 2024-10-17
[BUG] Make `model` required for `metadata.yaml` files


marc-white:
# Describe the bug

The current Builders/Translators seem to be unable to cope if the `metadata.yaml` for a new experiment is lacking a value for `model`. However, `model` is not defined to be a required element of the `metadata.yaml` files ([see the docs](https://access-nri-intake-catalog.readthedocs.io/en/latest/management/building.html#metadata-yaml-files)).

# To Reproduce

Attempt to do a test catalog build of the MOM6 data using the current HEAD of branch `175-data-request` etc. The following is reported out of the PBS job handler:

```
Successfully wrote ESM catalog json file to: file:///scratch/tm70/mcw120/intake-cat-test/v0.1.4/source/OM4_025.JRA_RYF.json
Successfully wrote ESM catalog json file to: file:///scratch/tm70/mcw120/intake-cat-test/v0.1.4/source/panant-01-hycom1-v13.json
Traceback (most recent call last):
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.6/bin/catalog-build", line 10, in <module>
    sys.exit(build())
  File "/home/120/mcw120/access-nri/access-nri-intake-catalog/src/access_nri_intake/cli.py", line 194, in build
    getattr(cm, method)(**args)
  File "/home/120/mcw120/access-nri/access-nri-intake-catalog/src/access_nri_intake/catalog/manager.py", line 127, in build_esm
    self._add()
  File "/home/120/mcw120/access-nri/access-nri-intake-catalog/src/access_nri_intake/catalog/manager.py", line 204, in _add
    self.dfcat.add(self.source, row.to_dict(), overwrite=overwrite)
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.6/lib/python3.10/site-packages/intake_dataframe_catalog/core.py", line 296, in add
    raise DfFileCatalogError(
intake_dataframe_catalog.core.DfFileCatalogError: Cannot add entry with iterable metadata columns: ['realm', 'frequency', 'variable'] to dataframe catalog with iterable metadata columns: ['model', 'realm', 'frequency', 'variable'].  Please ensure that metadata entries are consistent.
```

Given the `metadata.yaml` for these experiments had no `model` set, none was added to the constituent Intake-ESM catalogues. My attempts using the `Mom6Translator` to hard-code in a `model` value for these experiments is causing an inconsistency.

# Additional context

[...]

PR #255, opened 2024-11-15
223-model-required

charles-turner-1:
# Changes
- I fixed up a couple of tangentially related bits in the translators - looks like some WIP got accidentally merged in the @tuplify_series decorator & moved that to a separate file.
- Added more thorough error handling, checking for specific data issues & pointing to #223 is it appears to be related.
- Tests for error handling. 
   - The tests use mock raises - I'd really rather avoid this if possible & pass data broken in the specific expected ways, but it was sinking too much time so I'll raise it as a new issue & come back to check.

### DIFF
--- a/src/access_nri_intake/catalog/manager.py
+++ b/src/access_nri_intake/catalog/manager.py
@@ -203,14 +203,14 @@ class CatalogManager:
                 self.dfcat.add(self.source, row.to_dict(), overwrite=overwrite)
             except DfFileCatalogError as exc:
                 # If we have 'model' in the error message, it likely relates to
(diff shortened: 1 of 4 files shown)


charles-turner-1:
Will require [intake-dataframe-catalog](https://github.com/ACCESS-NRI/intake-dataframe-catalog/pull/71) too in order to work properly.

rbeucher (approved):

rbeucher:
Should we try to fix the coverage?

charles-turner-1:
Yeah, it's just a straightforward catch all that needs an additional test - should take 5 minutes, I'll do it first thing Monday.

PR #266, opened 2024-11-20
153-End to End build test.

### LINKED ISSUE #153: Intake Take2
[Intake Take2](https://intake.readthedocs.io/en/latest/index.html) is currently under development. It is a complete rewrite of Intake that aims "to be largely backward compatible with pre-V2 Intake sources and catalogs." However, Intake-ESM, which the access-nri-intake-catalog is built on, is a somewhat unusual application of Intake. At this point, it's not clear that there will be backwards compatibility for our application. This said, some of the new features promised by Intake Take2 may allow for a newer and better Intake-ESM, but this will obviously be a lot of work.

Strategy for now is to pin to Intake v1 and keep an eye on Intake Take2 progression/developments.

charles-turner-1:
Closes #153.

This contains an end to end build test, where a subset of the catalog is built, and queries executed against it.

- All queries on the subset catalog have been tested for correctness against the production catalog.
- I've used CMIP5 & a single OM2 experiment to test against.
- I've added a few minor refactors to the `clip.py` module to make testing easier.

@rbeucher can you link me to the Gadi SSH workflows you showed me the other day & I'll create a trigger for this based off those.


src/access_nri_intake/cli.py
```diff
@@ -87,18 +94,19 @@ def _check_build_args(args_list):
 
     if len(names) != len(set(names)):
         seen = set()
-        dupes = [name for name in names if name in seen or seen.add(name)]
+        dupes = [name for name in names if name in seen or seen.add(name)]  # type: ignore
+        # seen.add(name) returns None & so is always Falsey - so what is it doing?
```
marc-white, line 98: Please tell me `Falsey` is a British programming idiom :P

marc-white: Seriously though, I think the `or` is being used as a sneaky way to avoid an `if else` statement to add things to the "I've now seen this name" set (`seen`).

It's going in order through the `names` list. The `or` is evaluated left to right, so:

- If the name has been seen earlier in the list (i.e., `name in seen == True`), then it gets added to `dupes`. The `seen.add(name)` is never reached.
- If `name in seen == False`, then it moves on to the second part of the `or` statement, which adds `name` to the `seen` set. You're right that it's always False, but that's the point - it's `False` on _this_ pass, but now that it's been added to the `seen` set, if the same name recurs on a future pass, it is now known to be a duplicate.

(Interestingly, `False or None` evaluates to `None`, not `False`, but for this list comprehension I'm assuming anything not `True` is close enough to `False` to not care.)

```ipython
In [13]: (False or None) is None
Out[13]: True

In [14]: None or False
Out[14]: False
```

marc-white: Also, if a duplicated is repeated `N` times, then it will appear `(N-1)` times in the error message below - not sure if that's a bug or feature.

e2e/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+from pytest import fixture
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+
+def _get_xfail():
+    """
+    Get the XFAILS environment variable. We use a default of 1, indicating we expect
+    to add xfail marker to `test_parse_access_ncfile[AccessOm2Builder-access-om2/output000/ocean/ocean_grid.nc-expected0-True]`
+    unless specified.
```
marc-white, line 18: Probably should be updated to reference which test a value of `1` will point to.

e2e/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+@fixture(scope="session")
+def v_num():
+    return datetime.now().strftime("v%Y-%m-%d")
+
+
+def pytest_collection_modifyitems(config, items):
+    """
+    This function is called by pytest to modify the items collected during test
+    collection. We use it here to mark the xfail tests in
+    test_builders::test_parse_access_ncfile when we check the file contents & to
```
marc-white, line 57: As above

e2e/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+    ensure we correctly get xfails if we don't have cordinate discovery enabled
+    in intake-esm.
+    """
+    for item in items:
+        if (
+            item.name
+            in (
+                "test_parse_access_ncfile[AccessOm2Builder-access-om2/output000/ocean/ocean_grid.nc-expected0-True]",
```
marc-white, line 65: As above

e2e/test_end_to_end.py
```diff
@@ -0,0 +1,2517 @@
...
+            print(f"{indent}├── {item}")
+
+
+@pytest.fixture(scope="session")
+def current_catalog():
+    """
+    Return the current catalog as an intake catalog.
+    """
+    metacat = intake.cat.access_nri
+    yield metacat
```
marc-white, line 44: What's the idea behind `yield`ing here instead of `return`ing? (Not an issue, just for my own edification.)

e2e/test_end_to_end.py
```diff
@@ -0,0 +1,2517 @@
...
+
+@pytest.mark.parametrize(
+    "name",
+    [
+        "cmip5_al33",
+        "1deg_jra55_ryf9091_gadi",
+    ],
+)
+def test_datastore_found(metacat, name):
+    breakpoint()
```
marc-white, line 90: I presume this needs to be removed in production?

marc-white (commented):
At the risk of causing the build time to blow out, is it worth making sure each `Builder` is represented in the E2E test?

marc-white:
A couple of other things:

- I presume the `e2e` directory is outside the `tests` directory so that it can have a separate `conftest.py`. However, I think that should still work as expected if `e2e` is a subdirectory of `tests`. If that's not the case, could `e2e` be renamed to something like `tests_e2e` so we have better visibility that it's a test suite, not code?
- I think it would be nice if the standard `pytest` call **doesn't** try to execute the E2E tests. I think there's ways that a `skipif` can be used to make this happen, e.g.: [https://stackoverflow.com/questions/33084190/default-skip-test-unless-command-line-parameter-present-in-py-test](https://stackoverflow.com/questions/33084190/default-skip-test-unless-command-line-parameter-present-in-py-test)

rbeucher:
Hi @charles-turner-1,

Regarding the Gadi deployment actions, the workflow file can be found [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/main/.github/workflows/gadi_deploy.yml).

This workflow triggers the `build.sh` PBS script located in the `admin/access-nri-intake-catalog/bin` folder.

I’ve deployed the current main branch to the conda/access-med-0.10 environment. Fingers crossed, it works as expected!

rbeucher:
Just tested with a subset that only includes CMIP5 and it worked fine:
`https://github.com/ACCESS-NRI/access-nri-intake-catalog/actions/runs/11945531317`
We should have a way to trigger a build test from GitHub.

src/access_nri_intake/cli.py
charles-turner-1: Funnily enough, I think `Falsey` is pretty common worldwide - just googled it and it looks like it mostly crops up in javascript land though.

More importantly though, I had completely forgotten that the `or` will get executed left -> right & that `seen.add()` will still evaluate, even though it returns None.

I think the `False or None == None` and `None or False == False` thing is the result of the same trick - `None` and `False` are both falsey, so we wind up getting the second value in either case. Makes perfect sense, but completely counterintuitive.

Either way, I'll remove the unnecessary comment & see if I can strip it down to 1 error message, not N-1.

e2e/conftest.py
charles-turner-1: Good spot

charles-turner-1:
> A couple of other things:
> 
> * I presume the `e2e` directory is outside the `tests` directory so that it can have a separate `conftest.py`. However, I think that should still work as expected if `e2e` is a subdirectory of `tests`. If that's not the case, could `e2e` be renamed to something like `tests_e2e` so we have better visibility that it's a test suite, not code?
> * I think it would be nice if the standard `pytest` call **doesn't** try to execute the E2E tests. I think there's ways that a `skipif` can be used to make this happen, e.g.: https://stackoverflow.com/questions/33084190/default-skip-test-unless-command-line-parameter-present-in-py-test

I virtually always run pytest using `pytest tests` - hence sticking it in a separate E2E folder & adding that to the `.github_worfkows/ci.yml`. Funnily enough, I originally called it `e2e`, not `test_e2e` or similar to avoid pytest discovering & running it erroneously. Obviously, that didn't work (see 1.)

I think both your suggestions would make this much cleaner.

charles-turner-1:
I've done a bit more looking to this - frustratingly, looks like there might be some complications into the pytest `skipif` decorator not working on fixtures. Hopefully this should have little/no impact, but it's gonna take a touch more digging to make sure we can merge the confests & not cause any issues.

tests/e2e/test_end_to_end.py
```diff
@@ -0,0 +1,2502 @@
...
+import numpy as np
+import pytest
+
+from access_nri_intake.cli import build
+
+from .conftest import here
+
+e2e = pytest.mark.skipif(
+    "not config.getoption('--e2e')",
+)
```
marc-white, line 13: Idle curiosity - does this cause the option to get auto-documented somewhere, or do we need to remember to do that manually?

tests/e2e/test_end_to_end.py
```diff
@@ -0,0 +1,2502 @@
...
+@pytest.fixture(scope="module")
+def metacat(BASE_DIR, v_num):
+    # Build our subset of the catalog. This should take ~2 minutes with the PBS
+    # flags in build_subset.sh
+    print(f"Building the catalog subset & writing to {BASE_DIR}")
+    print(f"Version number: {v_num}")
+    build(
+        [
+            f"{here}/configs/cmip5.yaml",
+            f"{here}/configs/access-om2.yaml",
```
marc-white, line 36: Two separate-but-related points:
- If `here` is being referenced directly, should it be a fixture like, e.g., `test_data` is?
- I think it's generally safer to build file paths using `os` rather than direct string building.

tests/e2e/test_end_to_end.py
charles-turner-1: Turns out it does - `pytest --help` returns 
```bash
...

Custom options:
  --e2e                 Run end-to-end tests

...
```

Pleasantly surprised by that one - but I'll add it to the docs too.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,36 @@
...
+            export PROJECT=xp65_w
+      - name: Run end-to-end tests
+        uses: appleboy/ssh-action@v1.1.0
+        with:
+          host: gadi.nci.org.au
+          username: ${{secrets.GADI_USER}}
+          key: ${{secrets.DEPLOY_KEY}}
+          script: |
+            cd ${{secrets.GADI_REPO_PATH}}
+            qsub bin/build_all.sh
```
charles-turner-1, line 34: This should be a pytest script

marc-white (approved):
I think we're good to go now, merge when ready (and if you're satisfied the E2E testing is all set up correctly on the GitHub end).

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,36 @@
...
+        description: 'Release version'     
+        required: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        ### Latest at time of writing
+        uses: actions/checkout@v4.2.2
```
rbeucher, line 15: I would just assume we have a copy on Gadi for now. You can delete the checkout and the rsync.

rbeucher (approved):

PR #268, opened 2024-11-25
Test suite fixes (CLI locations & metadata outputs)

### LINKED ISSUE #265: [BUG] CLI tests will fail if not executed from top-level directory
# Describe the bug

If the test suite is executed via `pytest` from anywhere other than the top-level directory of the package, all of the `test_cli.py` tests will error out with `FileNotFoundError`:

```term
FAILED test_cli.py::test_build[v2024-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build[2024-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_bad_metadata - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata-bad.yaml'
FAILED test_cli.py::test_build_repeat_nochange - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_adddata - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data[v2001-01-01-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data[None-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data[v2001-01-01-None] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data_existing_old_cat[v2001-01-01-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data_existing_old_cat[None-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_existing_data_existing_old_cat[v2001-01-01-None] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_renamecatalogyaml[v2001-01-01-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_renamecatalogyaml[None-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_renamecatalogyaml[v2001-01-01-None] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct[v2001-01-01-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct[None-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct[v2001-01-01-None] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct_multivers[v2001-01-01-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct_multivers[None-v2099-01-01] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
FAILED test_cli.py::test_build_repeat_altercatalogstruct_multivers[v2001-01-01-None] - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/access-om2/metadata.yaml'
```

# To Reproduce

Attempt to run `pytest` from, e.g., inside the `tests` directory of the package, rather than at the top level.

# Additional context

I think the issue is coming from the paths within the test catalog YAML configs being hard-coded with respect to the top-level package directory, e.g., in `tests/data/config/access-om2.yaml`:

```yaml
builder: AccessOm2Builder

translator: DefaultTranslator

sources:

  - path:
      - tests/data/access-om2
    metadata_yaml: tests/data/access-om2/metadata.yaml
```

I'm not sure how to fix this yet - we'd need to intercept the read of these YAML files and prepend the absolute test path somehow (via `mock` or otherwise).

marc-white:
Fixes #265 (plus another bug found during testing).

This PR makes two improvements to the test suite:

1. The success/failure of the `test_cli.py` suite is now no longer dependent on running the tests from the top of the code directory tree. This was done by adding a `--data_base_path` option to the build script (usually the no-op value `""` in production), which allows for relative paths in configuration YAMLs to be resolved into absolute paths. (This may also have future uses in production, e.g. if a future catalog has a consistent data structure, but is being moved between different file system root paths.)
2. While I was testing this, I discovered the that `metadata-template` test was dumping a template YAML file into the PWD, which had accidentally been committed to the repo at some point. That stray file has now been removed, and `metadata-template` has been updated to allow it to send the template to an arbitrary location (by default, the PWD).


src/access_nri_intake/cli.py
```diff
@@ -27,7 +27,7 @@ class MetadataCheckError(Exception):
     pass
 
 
-def _parse_build_inputs(config_yamls, build_path):
+def _parse_build_inputs(config_yamls, build_path, data_path):
```
charles-turner-1, line 30: I think it might make more sense to rename `data_path` => `data_base_path` here? As far as I can see it's only called once & uses `data_base_path` in that call - might be easier to track in future keeping that variable name?

src/access_nri_intake/cli.py
```diff
@@ -54,10 +54,14 @@ def _parse_build_inputs(config_yamls, build_path):
         for kwargs in sources:
             source_args = config_args
 
-            source_args["path"] = kwargs.pop("path")
+            source_args["path"] = [
+                os.path.join(data_path, _) for _ in kwargs.pop("path")
```
charles-turner-1, line 58: Not specific to this but in general we seem to be using pathlib.Path and os.path throughout this module (and package). I think we should probably move over to doing everything with pathlib.Path?

src/access_nri_intake/cli.py
```diff
@@ -141,6 +145,16 @@ def build():
         ),
     )
 
+    parser.add_argument(
+        "--data_base_path",
+        type=str,
+        default="",
```
charles-turner-1, line 151: Does it make more sense to use a base path of `'./'` rather than an empty default? I've tried to follow through the use of data_base_path & I really can't tell - `'./'` intuitively feels cleaner to me, but I feel like there might be a good reason why you haven't done that here I'm not seeing?

tests/test_cli.py
```diff
@@ -288,6 +294,7 @@ def test_build_repeat_nochange(mockargs, get_catalog_fp, test_data):
         ],
         build_base_path=tempfile.TemporaryDirectory().name,  # Use pytest fixture here?
```
charles-turner-1, line 295: Again unrelated to this PR but I think using the tempfile module like this won't tear down the temporary directory when we're done with it properly & we might want to use a fixture (I'm pretty sure you left the comment noting fixtures).

tests/test_cli.py
```diff
@@ -958,4 +978,7 @@ def test_metadata_validate_no_file(mockargs):
 
 
 def test_metadata_template():
-    metadata_template()
+    loc = tempfile.TemporaryDirectory()
```
charles-turner-1, line 981: As above about tempfile. I've added a temporary filesystem fixture in the E2E stuff, so I think we can come back and address this later TBH - I think we'll want to merge both in before addressing tempfile stuff to avoid effort duplication.

src/access_nri_intake/cli.py
marc-white: I chose it because it no-ops `os.path.join`:

```ipython
In [2]: os.path.join("", "/my/existing/path")
Out[2]: '/my/existing/path'
```

`./` does do the same, however:

```ipython
In [3]: os.path.join("./", "/my/existing/path")
Out[3]: '/my/existing/path'
```

However, the output does differ if the path you join to is relative (although only syntactically, the effective output is identical):

```ipython
In [4]: os.path.join("./", "my/existing/path")
Out[4]: './my/existing/path'

In [5]: os.path.join("", "my/existing/path")
Out[5]: 'my/existing/path'
```

tests/test_cli.py
marc-white: Actually I didn't, I had duplicated the line from something someone else had done (I assume @dougiesquire), which explains why it's there.

The issue is that I often need multiple, known-to-be-different temporary directories to be able to run many of the tests correctly - is there a way to set up some sort of fixture 'factory' for this?

tests/test_cli.py
marc-white: OK let's address the tempfile/tempdir things in a new issue/PR, but do that as a priority - we don't want to be flooding test systems accidentally.

tests/test_cli.py
marc-white: Just saw your other comment about fixturing in the E2E branch, so we'll hold this over for a new issue/PR.

src/access_nri_intake/cli.py
charles-turner-1: Cool - I've done a bit of looking through the docs & it looks like functionally they're equivalent in every way - if we prepend a relative path to an absolute path, the relative path will be ignored.

I think it would make more sense to use `"./"` just for consistency with `--build_base_path` & `--catalog_base_path`, but I don't think it's crucial.

charles-turner-1 (approved):
I think I'd prefer `""` => `"./"` (somewhere in comments) but otherwise this looks ready to merge.

rbeucher (approved):

src/access_nri_intake/cli.py
marc-white: I started playing with this, and it might be a bit of extra work - there's quite the mix of `Path` objects and path strings flying around. Plus, I see you've already made an issue for this, so I'll mark this comment as resolved and we can fix that up later.

PR #276, opened 2024-12-02
Switch to correct tmp_path usage in tests

### LINKED ISSUE #270: [Potential BUG] Teardown of temporary file structures in tests may not be complete
# Describe the bug

There are some tests in `tests/test_cli.py` which use temporary directories: eg [link](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/2fdc8cca661671d9eb9b621bfc021da61299f51f/tests/test_cli.py#L235). 

The tempfile module can be a bit finnicky when not using a context manager & I think there's the possibility that we're not cleaning up out temporary directories in an ideal fashion.

PR #266 (end to end build test) uses pytest's built in fixtures to handle the temporary filesystem stuff, and I think it would be ideal to move to using this as soon as we can.

# Additional context

Noted during #268 .

marc-white:
Fixes #270 .

The `test_cli` suite previously used `tempfile.TemporaryDirectory` to create test spaces. This likely wasn't being torn down properly, so those have all been switched to use the `pytest` `tmp_path` fixture.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -5,7 +5,6 @@ import argparse
 import glob
 import os
 import shutil
-import tempfile
 from pathlib import Path
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


charles-turner-1 (approved):
Looks good to me. 

I've also run the test suite on this with [this plugin](https://pypi.org/project/pytest-random-order/), just for completeness, and everything looks good.

Ready to merge.

PR #293, opened 2024-12-04
292 flaky tests

### LINKED ISSUE #292: [BUG] `tests/test_builders.py::test_builder_build` Flaky
# Describe the bug

We were getting some weird test behaviour in #291 - and after a bit of digging, I noticed we are getting a [CI failure on main](https://github.com/ACCESS-NRI/access-nri-intake-catalog/actions/runs/12153053028).

I couldn't figure out exactly why, because I couldn't reproduce the failure locally - so I deleted & recloned the repo, and which reproduced the failure. Running the test suite a couple more times resulted in the test suite passing, with no apparent changes to the file structure of the repo.

# To Reproduce

1. Start with a fresh install - either
```bash
$ git clone https://github.com/ACCESS-NRI/access-nri-intake-catalog/
$ conda env create -f environment-3.13.yml --name access-nri-intake 
$ pip install -e .
```
 or 
```bash
$ git clone https://github.com/ACCESS-NRI/access-nri-intake-catalog/ 
$ conda env create -f environment-3.13.yml --name access-nri-intake
$ pip install .
``` 

3. Run `pytest tests/test_builders.py::test_builder_build` 
4. You should get 
```python
>       assert len(builder.df) == num_valid_assets
E       AssertionError: assert 26 == 27
```

5.  Repeat 2 twice. The error/failure should occur again on the second run, but not the 3rd.

## Additional info

On python3.12 (change environment in step 1), it took four runs - starting with 
```python
E       AssertionError: assert 25 == 27
```
and then incrementing to 26, 26, and then finally 27.

It looks to me like we might have some test pollution going on here, but I'm not quite sure what.

Potentially related to #270 and #276?

charles-turner-1:
Reverting addition of MOM6 stuff seemed to fix tests (CI :X: on initial commit is due to codecov, tests all passed).

- Note that unreverting (revert revert) tests actually passed on Python3.10, but failed on 3.13 as CI shut everything down after 3.13 failure.

 Closes #292 ... when we understand what's going on.


charles-turner-1:
Note - I only opened a PR to trigger the CI runs.

charles-turner-1:
This should fix all the MOM6 issues.

- I'm not entirely happy with the fix - it's more of a sidestep than anything & we will need to take a closer look at the cftime dependency which is the source of issue, but other than fixing coverage this should resolve all the problems we were having with Mom6 stuff.

TLDR; the time bounds in the mom6 model cause a lot of issues.

src/access_nri_intake/source/builders.py
```diff
@@ -543,8 +544,76 @@ def parser(cls, file):
...
 
         except Exception:
+            if "NotImplementedError" in traceback.format_exc():
+                raise NotImplementedError(f"Ready to implement for {file}")
             return {INVALID_ASSET: file, TRACEBACK: traceback.format_exc()}
 
+    @classmethod
+    def parse_access_ncfile(
+        cls, file: str, time_dim: str = "time"
+    ) -> _AccessNCFileInfo:
```
marc-white, line 554: This looks pretty similar to the standard `parse_access_ncfile` to me, just with a change to the TimeParser object being used. Can the time parser be made an optional kwarg of the standard `parse_access_file` and we toss this version, or is there something else I've missed?

charles-turner-1: Bear with me - I'm gonna rename a few slightly misnamed functions & hopefully the intention should become clearer

charles-turner-1:
Should be ready for review. There's one line that just needs mocking - we need to patch in an empty list here, but I'm struggling to work out why mocking a dataclass attribute won't work.
```python
595        if not dvars.variable_list:
596           raise EmptyFileError("This file contains no variables")
```

src/access_nri_intake/source/builders.py
```diff
@@ -285,9 +286,7 @@ def parse_access_filename(
         return file_id, timestamp, frequency
 
     @classmethod
-    def parse_access_ncfile(
-        cls, file: str, time_dim: str = "time"
-    ) -> _AccessNCFileInfo:
+    def parse_ncfile(cls, file: str, time_dim: str = "time") -> _NCFileInfo:
         """
         Get Intake-ESM datastore entry info from an ACCESS netcdf file
```
marc-white, line 291: May need to alter this description

src/access_nri_intake/source/builders.py
```diff
@@ -545,6 +544,70 @@ def parser(cls, file):
         except Exception:
             return {INVALID_ASSET: file, TRACEBACK: traceback.format_exc()}
 
+    @classmethod
+    def parse_ncfile(cls, file: str, time_dim: str = "time") -> _NCFileInfo:
```
marc-white, line 548: This is still a line-for-line copy of `BaseBuilder.parse_ncfile`, just with the time parser object changed. Is it workable to add a `time_parser` kwarg to `BaseBuilder.parse_ncfile`, and just inherit from there?

If you want clearer separation, your other option might be to come up with a Mixin solution that defines `parse_ncfile` separately (which is workable, given it's a classmethod.)

marc-white:
Looks good to me, modulo my comments above.

src/access_nri_intake/source/builders.py
charles-turner-1: Ahh gotcha - yeah, the first solution makes a lot more sense than what I've got here.

charles-turner-1: Turned out to be easier from a Liskov perspective to just add the time parser as a class attribute - but I think the changes should address your points.

If you're happy, I think we should be ready to merge this in.

Issue #294, opened 2024-12-06
Improve catalog version reporting


marc-white:
# Is your feature request related to a problem? Please describe.

At present, the only way to know what catalog versions are available is to directly look at the directories in `/g/data/xp65/public/apps/access-nri-intake-catalog/`. This is, obviously, not a great solution.

# Describe the feature you'd like

There should be a function, available in Python, that lists all available catalog versions. 

The function should list:
- All versions available with the current `catalog.yaml` file;
- All versions available if a user copies one of the old `catalog.yaml` files to their home area.

# Describe alternatives you've considered

Nil.

# Additional context

This is handy for use with a post-catalog build test notebook (#279 ).

marc-white:
I'd forgotten that version reporting already exists (`access_nri_intake.utils.available_versions`), but it's pretty limited, can't handle outdated catalog YAMLs, etc. I'm updating it.

marc-white:
Re-opening this issue - still need to implement version reporting for outdated catalogs, accessible by out-of-date `catalog-<min.vers.>-<max.vers.>.yaml` files. (None of these exist yet, so not urgent.)

PR #295, opened 2024-12-09
Progress on catalog version reporting

marc-white:
Progress on #294 .

This PR updates the existing `available_versions` function to:

- Use `pathlib.Path` instead of the `os` module;
- Consider the min/max/default versions of the catalog specified in the `catalog.yaml` when reporting available versions;
- Mark the default version in the `pretty` printout return.

Tests have also been updated as required.

### DIFF
--- a/src/access_nri_intake/data/utils.py
+++ b/src/access_nri_intake/data/utils.py
@@ -12,7 +12,7 @@ from . import CATALOG_NAME_FORMAT
 CATALOG_PATH_REGEX = r"^(?P<rootpath>.*?)\{\{version\}\}.*?$"
 
 
-def _get_catalog_rp():
+def _get_catalog_root():
     """
     Get the catalog root path.
     """
@@ -35,7 +35,7 @@ def _get_catalog_rp():
         )
 
 
-def available_versions(pretty: bool = True):
+def available_versions(pretty: bool = True) -> list[str] | None:
     """
     Report the available versions of the `intake.cat.access_nri` catalog.
 
@@ -46,12 +46,12 @@ def available_versions(pretty: bool = True):
         (True, default), or to provide a list of version numbers only (False).
     """
     # Work out where the catalogs are stored
-    base_path = _get_catalog_rp()
+    base_path = _get_catalog_root()
 
     # Grab the extant catalog and work out its min and max versions
     try:
-        with open(get_catalog_fp()) as cat:
-            cat_yaml = yaml.safe_load(cat)
+        with open(get_catalog_fp()) as cat_file:
+            cat_yaml = yaml.safe_load(cat_file)
             vers_min = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["min"]
             vers_max = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["max"]
             vers_def = cat_yaml["sources"]["access_nri"]["parameters"]["version"][
@@ -64,11 +64,14 @@ def available_versions(pretty: bool = True):
 
     # Grab all the catalog names
     cats = [
-        d.name
-        for d in base_path.iterdir()
-        if re.search(CATALOG_NAME_FORMAT, d.name)
-        and d.is_dir()
-        and ((d.name >= vers_min and d.name <= vers_max) or d.name == vers_def)
+        dir_path.name
+        for dir_path in base_path.iterdir()
+        if re.search(CATALOG_NAME_FORMAT, dir_path.name)
+        and dir_path.is_dir()
+        and (
+            (dir_path.name >= vers_min and dir_path.name <= vers_max)
+            or dir_path.name == vers_def
+        )
     ]
     cats.sort(reverse=True)
 
@@ -84,6 +87,6 @@ def available_versions(pretty: bool = True):
             if c == vers_def:
                 c += "*"
             print(c)
-        return
+        return None
 
     return cats
--- a/src/access_nri_intake/data/utils.py
+++ b/src/access_nri_intake/data/utils.py
@@ -6,8 +6,8 @@ import re
 
 import yaml
 
+from ...access_nri_intake import CATALOG_VERSION_REGEX
 from ..utils import get_catalog_fp
-from . import CATALOG_NAME_FORMAT
 
 CATALOG_PATH_REGEX = r"^(?P<rootpath>.*?)\{\{version\}\}.*?$"
 
@@ -49,7 +49,7 @@ def available_versions(pretty: bool = True):
     base_path = _get_catalog_rp()
 
     # Grab all the catalog names
-    cats = [d for d in os.listdir(base_path) if re.search(CATALOG_NAME_FORMAT, d)]
+    cats = [d for d in os.listdir(base_path) if re.search(CATALOG_VERSION_REGEX, d)]
     cats.sort(reverse=True)
 
     # Find all the symlinked versions
--- a/src/access_nri_intake/data/utils.py
+++ b/src/access_nri_intake/data/utils.py
@@ -82,6 +82,8 @@ def available_versions(pretty: bool = True):
         for c in cats:
             if c in symlink_targets.keys():
                 c += f"(-->{symlink_targets[c]})"
+            if c == vers_def:
+                c += "*"
             print(c)
         return
 
--- a/src/access_nri_intake/data/utils.py
+++ b/src/access_nri_intake/data/utils.py
@@ -3,6 +3,7 @@
 
 import os
 import re
+from pathlib import Path
 
 import yaml
 
@@ -28,7 +29,7 @@ def _get_catalog_rp():
 
     match = re.match(CATALOG_PATH_REGEX, catalog_fp)
     try:
-        return match.group("rootpath")
+        return Path(match.group("rootpath"))
     except AttributeError:  # Match failed
         raise RuntimeError(
             f"Catalog metadata {get_catalog_fp()} contains unexpected catalog filepath: {catalog_fp}"
--- a/src/access_nri_intake/data/utils.py
+++ b/src/access_nri_intake/data/utils.py
@@ -6,8 +6,8 @@ import re
 
 import yaml
 
-from ...access_nri_intake import CATALOG_VERSION_REGEX
 from ..utils import get_catalog_fp
+from . import CATALOG_NAME_FORMAT
 
 CATALOG_PATH_REGEX = r"^(?P<rootpath>.*?)\{\{version\}\}.*?$"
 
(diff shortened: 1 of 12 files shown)


marc-white:
I added more tests, but codecov still thinks I'm not testing L60-63 of `data.utils.available_versions`.

charles-turner-1:
I'll take a look at this first thing tomorrow morning.

src/access_nri_intake/data/utils.py
```diff
@@ -48,21 +48,41 @@ def available_versions(pretty: bool = True):
...
+            ]
+    except FileNotFoundError:
+        raise FileNotFoundError(f"Unable to find catalog at {get_catalog_fp()}")
+    except KeyError:
+        raise RuntimeError(f"Catalog at {get_catalog_fp()} not correctly formatted")
+
     # Grab all the catalog names
-    cats = [d for d in os.listdir(base_path) if re.search(CATALOG_NAME_FORMAT, d)]
+    cats = [
+        d.name
```
charles-turner-1, line 67: Minor nitpick but I find `for dir in ..` rather than `for d in ...` more readable - the list comprehension spans a few lines already so no harm in readability from the extra characters.

src/access_nri_intake/data/utils.py
```diff
@@ -48,21 +48,41 @@ def available_versions(pretty: bool = True):
...
+    symlinks = [s for s in cats if (Path(base_path) / s).is_symlink()]
 
-    symlink_targets = {
-        s: os.path.basename(os.readlink(os.path.join(base_path, s))) for s in symlinks
-    }
+    symlink_targets = {s: (base_path / s).readlink().name for s in symlinks}
 
     if pretty:
-        for i, c in enumerate(cats):
+        for c in cats:
```
charles-turner-1, line 81: As above about single char varnames - `for cat in cats` easier to read?

src/access_nri_intake/data/utils.py
charles-turner-1: NB. I only find this holds true with multi line expressions - maybe I need to get my eyes checked.

src/access_nri_intake/data/utils.py
```diff
@@ -48,21 +48,41 @@ def available_versions(pretty: bool = True):
     # Work out where the catalogs are stored
     base_path = _get_catalog_rp()
```
charles-turner-1, line 49: Not immediately related to the changes on this PR but I only just realised we have a function called `get_catalog_fp` and another called `_get_catalog_rp` - I think we could make these more obviously distinct?

charles-turner-1: Spent about 10 minutes reading this file before I even noticed they were two separate functions

src/access_nri_intake/data/utils.py
```diff
@@ -48,21 +48,41 @@ def available_versions(pretty: bool = True):
...
 
     if pretty:
-        for i, c in enumerate(cats):
+        for c in cats:
             if c in symlink_targets.keys():
                 c += f"(-->{symlink_targets[c]})"
+            if c == vers_def:
+                c += "*"
             print(c)
         return
```
charles-turner-1, line 87: Can we update `return` => `return None` and add a type hint on the function definition?

charles-turner-1 (commented):
I suspect that codecov might be getting upset because of the way the mocks are defined?
eg. test_data:
```python

@mock.patch("access_nri_intake.data.utils.get_catalog_fp")
-77 def test_available_versions_no_catalog(mock_get_catalog_fp):
-78    mock_get_catalog_fp.return_value = "/this/is/not/real.yaml"
-79    with pytest.raises(FileNotFoundError):
-80        available_versions()
```

```python
+ @mock.patch("access_nri_intake.data.utils.get_catalog_fp")
+ def test_available_versions_no_catalog(mock_get_catalog_fp):
+    mock_get_catalog_fp.return_value = "/this/is/not/real.yaml"
+    with pytest.raises(FileNotFoundError):
+        available_versions()
```

Intellisense is complaining to me that the `mock_get_catalog_fp` fixture is unused - maybe that's what's throwing codecov off the trail? 

Otherwise, a few minor nitpicks (I think I'm a bit jumpy after letting the overflow bug into main) but otherwise looks good.

src/access_nri_intake/data/utils.py
marc-white: Yeah, that's fair. Just for clarity:

- `get_catalog_fp` returns the full file path to the `catalog.yaml` file;
- `_get_catalog_rp` reads the `catalog.yaml`, and computes the root path (`rp`) to the catalog data by stripping off the `{{version}}/metacatalog.yaml` from the data path therein.

src/access_nri_intake/data/utils.py
marc-white: Fair comment, although is `dir` a dangerous/possibly protected name? Something like `dir_path` is probably safer (and makes it clearer that `d`/`dir`/`dir_path` is a `Path` object).

charles-turner-1: I *think* that dir is a soft keyword, a la match, but yeah, that is a good point - `dir_path` seems sensible.

src/access_nri_intake/data/utils.py
charles-turner-1: Makes sense. Maybe `_get_catalog_rp` => `_get_catalog_root` given it's a local function & so easier to track down the definition/meaning?

src/access_nri_intake/data/utils.py
```diff
@@ -48,21 +48,41 @@ def available_versions(pretty: bool = True):
...
```
marc-white, line 88: @charles-turner-1 having made the suggested changes, I'm now getting a `mypy` pre-commit error on this line:

```
src/access_nri_intake/data/utils.py:86: error: Unsupported left operand type for + ("TextIOWrapper[_WrappedBuffer]")  [operator]
src/access_nri_intake/data/utils.py:88: error: Unsupported left operand type for + ("TextIOWrapper[_WrappedBuffer]")  [operator]
```

Any hints? I'm not sure what's caused it to suddenly appear... the tests all pass fine.

charles-turner-1: What does your implementation look like? The following passes mypy just fine for me:

```python

def available_versions(pretty: bool = True) -> list[str] | None:
    """
    Report the available versions of the `intake.cat.access_nri` catalog.

    Parameters
    ---------
    pretty : bool, optional
        Defines whether to return a pretty print-out of the available versions
        (True, default), or to provide a list of version numbers only (False).
    """
    # Work out where the catalogs are stored
    base_path = _get_catalog_rp()

    # Grab the extant catalog and work out its min and max versions
    try:
        with open(get_catalog_fp()) as cat:
            cat_yaml = yaml.safe_load(cat)
            vers_min = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["min"]
            vers_max = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["max"]
            vers_def = cat_yaml["sources"]["access_nri"]["parameters"]["version"][
                "default"
            ]
    except FileNotFoundError:
        raise FileNotFoundError(f"Unable to find catalog at {get_catalog_fp()}")
    except KeyError:
        raise RuntimeError(f"Catalog at {get_catalog_fp()} not correctly formatted")

    # Grab all the catalog names
    cats = [
        d.name
        for d in base_path.iterdir()
        if re.search(CATALOG_NAME_FORMAT, d.name)
        and d.is_dir()
        and ((d.name >= vers_min and d.name <= vers_max) or d.name == vers_def)
    ]
    cats.sort(reverse=True)

    # Find all the symlinked versions
    symlinks = [s for s in cats if (Path(base_path) / s).is_symlink()]

    symlink_targets = {s: (base_path / s).readlink().name for s in symlinks}

    if pretty:
        for c in cats:
            if c in symlink_targets.keys():
                c += f"(-->{symlink_targets[c]})"
            if c == vers_def:
                c += "*"
            print(c)
        return None

    return cats
```

marc-white: I think going `c` --> `cat` broke/confused things... how odd.

marc-white, line 88: Hmmm I can't see any differences there, can you try this? (Note some variable name refactoring may be needed.)

```python
def available_versions(pretty: bool = True) -> list[str] | None:
    """
    Report the available versions of the `intake.cat.access_nri` catalog.

    Parameters
    ---------
    pretty : bool, optional
        Defines whether to return a pretty print-out of the available versions
        (True, default), or to provide a list of version numbers only (False).
    """
    # Work out where the catalogs are stored
    base_path = _get_catalog_root()

    # Grab the extant catalog and work out its min and max versions
    try:
        with open(get_catalog_fp()) as c:
            cat_yaml = yaml.safe_load(c)
            vers_min = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["min"]
            vers_max = cat_yaml["sources"]["access_nri"]["parameters"]["version"]["max"]
            vers_def = cat_yaml["sources"]["access_nri"]["parameters"]["version"][
                "default"
            ]
    except FileNotFoundError:
        raise FileNotFoundError(f"Unable to find catalog at {get_catalog_fp()}")
    except KeyError:
        raise RuntimeError(f"Catalog at {get_catalog_fp()} not correctly formatted")

    # Grab all the catalog names
    cats = [
        dir_path.name
        for dir_path in base_path.iterdir()
        if re.search(CATALOG_NAME_FORMAT, dir_path.name)
        and dir_path.is_dir()
        and (
            (dir_path.name >= vers_min and dir_path.name <= vers_max)
            or dir_path.name == vers_def
        )
    ]
    cats.sort(reverse=True)

    # Find all the symlinked versions
    symlinks = [s for s in cats if (Path(base_path) / s).is_symlink()]

    symlink_targets = {s: (base_path / s).readlink().name for s in symlinks}

    if pretty:
        for c in cats:
            if c in symlink_targets.keys():
                c += f"(-->{symlink_targets[c]})"
            if c == vers_def:
                c += "*"
            print(c)
        return None

    return cats
```

charles-turner-1: I'm pretty sure this is because of the repeated variable name:
```python
53   with open(get_catalog_fp()) as c:
54            cat_yaml = yaml.safe_load(c)
...
```
and
```python
84    if pretty:
85        for c in cats:
```

Mypy is inferring that these are the same variable - hence the error. If you change

```python
-        with open(get_catalog_fp()) as c:
-            cat_yaml = yaml.safe_load(c)

+          with open(get_catalog_fp()) as cat_file:
+            cat_yaml = yaml.safe_load(cat_file)
```
or similar it should fix it.

charles-turner-1: Alternatively, I think you can add a type hint comment - something like 
```python
        for c in cats: #type: string
```

marc-white: Yeah, it was the repeat variable name - got in OK now.

charles-turner-1:
Whoops, accidentally requested a review instead of giving one...

charles-turner-1 (approved):
Looks to me like codecov is talking nonsense, so I think this is ready to merge.

PR #298, opened 2024-12-10
297 translator errors

charles-turner-1:
- Added error handling 
- Draft for a decorator that should keep the error handling all in one place - having some issues getting the correct function name to pass through though.

### DIFF
--- a/src/access_nri_intake/catalog/utils.py
+++ b/src/access_nri_intake/catalog/utils.py
@@ -48,7 +48,7 @@ def trace_failure(func: Callable) -> Callable:
             return func(*args, **kwargs)
         except KeyError as exc:
             raise KeyError(
-                f"Unable to find {colname} column '{dispatch_key}' with translator {args[0].__class__.__name__}"
+                f"Unable to find {colname} column '{dispatch_key}' with translator '{args[0].__class__.__name__}'"
             ) from exc
 
     return wrapper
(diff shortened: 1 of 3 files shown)


src/access_nri_intake/catalog/utils.py
```diff
@@ -26,3 +27,28 @@ def wrapper(*args, **kwargs):
...
 
     return wrapper
+
+
+def trace_failure(func: Callable) -> Callable:
+    """
+    Decorator that wraps a function and prints a message if it raises an exception
+    """
+    func_name = func.__name__
+    colname = func_name[1:].split("_")[0]
```
marc-white, line 37: Exactly what issue are you having with getting the function name out?

charles-turner-1: `func.__name__` will return `wrapper` here - seems to be a common problem eg. https://stackoverflow.com/questions/50350379/getting-function-name-from-nested-decorator, but none of the solutions I've found seem to have solved it yet.

charles-turner-1: Fixed by changing the order of decoration & moving definitions in - about to tidy up & push.

marc-white: Cool, if you take it out of Draft I can approve it so it can go in

charles-turner-1:
Codecov is complaining that this decreases project coverage - but the coverage on the diff is 100% so...

marc-white:
> Codecov is complaining that this decreases project coverage - but the coverage on the diff is 100% so...

Yeah, I think we just ignore that.

marc-white (approved):
Looks good, approved to merge. I'll override the branch protection and merge it.

PR #304, opened 2024-12-11
199 cordex xfails

charles-turner-1:
- Add xfails on cordex translator (broken - see #199).
- Update storage flags, correct package dir
-  'Robustify' a couple of tests

PR #306, opened 2024-12-11
Consolidate on pathlib module

### LINKED ISSUE #271: Consolidation on use of pathlib module for filesystem operations, rather than pathlib & os.path
Currently, this package uses both `pathlib.Path` operations & `os.path.join` in a lot of places, with usage of both in plenty of files.

Pathlib has been the preferred way of doing things [since Python 3.4 ](https://stackoverflow.com/questions/67112343/pathlib-vs-os-path-join-in-python), and so we should think about moving over to using pathlib where possible.

Low priority.

marc-white:
Closes #271.

This PR aims to remove `os.path` dependency from the main package, replacing it with `pathlib.Path`.

An e2e test should be done on this branch before merging to confirm the build system is unaffected.

`os.path` is still used in `versioneer.py`, but I think that's code that's been generated externally and committed to our repo, so I haven't touched that.


charles-turner-1 (approved):
Looks good to me - as far as I can see you've caught them all:
```bash
$ rg --files-with-matches 'os\.path'
versioneer.py
tests/data/mom6/README
```

marc-white:
Running e2e tests needs to wait until #304 is merged in.

PR #312, opened 2024-12-17
311 telemetry

charles-turner-1:
Hooks in telemetry IPython extension upon accessing `intake.cat.access_nri` - and sends searches performed on either `DfFileCatalog` objects (ie. the metacatalog), or `esm_datastore` objects to a telemetry server.

Relies on [access-ipy-telemetry](https://github.com/ACCESS-NRI/access-ipy-telemetry)to provide the implementation of telemetry.


charles-turner-1:
Closing this as the testing infrastructure required is too different to the current setup & the functionality can be generalised - it's better off as it's own [package](https://github.com/ACCESS-NRI/access-ipy-telemetry).

charles-turner-1:
Tests failing due to [https://github.com/ACCESS-NRI/access-ipy-telemetry](access-ipy-telemetry) package not being available on PyPI yet - will need to upload to PyPI when I get my hands on the org credentials.

marc-white (commented):
I think the `git` dependency is breaking my editable install so I can't run the tests: 

```
Obtaining file:///Users/marc/Documents/ACCESS-NRI/access-nri-intake-catalog/access-nri-intake-catalog
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Requirement already satisfied: cftime in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (1.6.4)
Requirement already satisfied: ecgtools>=2023.7.13 in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (2024.7.31)
Requirement already satisfied: intake>=0.7.0 in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (0.7.0)
Requirement already satisfied: intake-dataframe-catalog>=0.2.4 in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (0.2.4)
Requirement already satisfied: intake-esm>=2023.11.10 in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (2024.2.6)
Requirement already satisfied: jsonschema in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (4.23.0)
Requirement already satisfied: pooch in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (1.8.2)
Requirement already satisfied: xarray in /Users/marc/anaconda3/envs/access-nri-intake-test/lib/python3.11/site-packages (from access_nri_intake==1.0.0+14.g9c67e7f) (2024.10.0)
INFO: pip is looking at multiple versions of access-nri-intake to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement access-ipy-telemetry>=0.1.0 (from access-nri-intake) (from versions: none)
ERROR: No matching distribution found for access-ipy-telemetry>=0.1.0
```

marc-white (commented):
My attempt to replicate the install is not going well:

```
pip install https://github.com/ACCESS-NRI/access-ipy-telemetry.git                                                                                                      1 ✘  4s  access-nri-intake-test Py  13:09:59 
Collecting https://github.com/ACCESS-NRI/access-ipy-telemetry.git
  Downloading https://github.com/ACCESS-NRI/access-ipy-telemetry.git
     - 388.3 kB 10.8 MB/s 0:00:00
  ERROR: Cannot unpack file /private/var/folders/93/b5zp1z_j4r35lgsbh92wkyy80000gp/T/pip-unpack-cstb5xru/access-ipy-telemetry.git (downloaded from /private/var/folders/93/b5zp1z_j4r35lgsbh92wkyy80000gp/T/pip-req-build-bes3kd5a, content-type: text/html; charset=utf-8); cannot detect archive format
ERROR: Cannot determine archive format of /private/var/folders/93/b5zp1z_j4r35lgsbh92wkyy80000gp/T/pip-req-build-bes3kd5a
```

charles-turner-1:
Weird... will look into it.

charles-turner-1:
Let's leave this for now - we'll need to add that package to PyPI at some point anyway, so might as well get that done and come back to it.

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "jsonschema",
     "pooch",
     "xarray",
+    "access-py-telemetry>=0.1.0",
```
marc-white, line 25: I can't seem to find this on conda.... will that break the conda package build?

src/access_nri_intake/data/__init__.py
```diff
@@ -21,3 +27,9 @@
         stacklevel=2,
     )
     data = intake.catalog.Catalog()
+finally:
+    cat_version = data._captured_init_kwargs.get("metadata", {}).get(
+        "version", "latest"
+    )  # Get the catalog version number and set it to "latest" if it can't be found
+    configure_telemetry(["--enable", "--silent"])
```
marc-white, line 34: Can a message be emitted here to warn the user that the telemetry has been turned on, and how to turn it back off if they so wish?

marc-white (changes requested):
All the code updates themselves look pretty straightforward, so no real comments there (modulo wondering what happens with conda builds).

However, I do have another concern about personal data management & tracking. I noticed on the `access_py_telemetry` docs that the telemetry is on by default on the conda environments. In that case, I think there need to be plenty of loud, obvious messages that come up saying that that's the case, and including the instructions for disabling it. I've also added a comment where the telemetry is (re-)enabled by this package.

pyproject.toml
charles-turner-1: Good point, I don't know the answer. I've put the package onto PyPI and set the CI environments to find it with pip. I'll find out what I need to do to add it to conda too for completeness.

src/access_nri_intake/data/__init__.py
charles-turner-1: N.B I am a bit of tinfoil hat and if I got a notice regarding telemetry and how to disable it, I would always do so. I'm assuming that is pretty common, but I don't know for sure

marc-white: Perhaps flash the message up at load time saying we're using telemetry, but leave the instructions/details to the docs? Then the true tin-foilers can look into how to disable it, but most might not bother?

charles-turner-1: Yeah, this might be the best option.

rbeucher: I agree with @charles-turner-1 I think if we flash messages every time a user loads the package, they will just deactivate telemetry and we will use important data. 

Any chance we could do that at install time?

Otherwise flagging it in the documentation is good enough for me.

pyproject.toml
marc-white: Did this end up getting covered?

charles-turner-1: Nah, I haven't got a chance yet - I'll take a look at it now.

src/access_nri_intake/data/__init__.py
marc-white: Where should we add in the documentation? The usage FAQs?

aidanheerdegen: We definitely want to let users know what we're collecting and why. For data users it's a very compelling case: we will rely on the telemetry to make important resource allocation decisions. If they turn off the telemetry we won't know they're using a dataset, which means it risks being removed/de-prioritised.

That sort of involved argument can only be made in the docs I think.

I wonder if there is a possibility that we could also give users some useful information from the telemetry so they would see some immediate value in it?

Like could they call a function that reports a summary of what they've accessed in a session? For their own interest?

Or do we need to worry about people gaming the system so the datasets they're interested in are well used?

charles-turner-1: > We definitely want to let users know what we're collecting and why. For data users it's a very compelling case: we will rely on the telemetry to make important resource allocation decisions. If they turn off the telemetry we won't know they're using a dataset, which means it risks being removed/de-prioritised.
> 
> That sort of involved argument can only be made in the docs I think.
> 
> I wonder if there is a possibility that we could also give users some useful information from the telemetry so they would see some immediate value in it?
> 
> Like could they call a function that reports a summary of what they've accessed in a session? For their own interest?

This is definitely doable & should be relatively straightforward.
> 
> Or do we need to worry about people gaming the system so the datasets they're interested in are well used?

My guess is probably not - I imagine that the amount of noisy signal generated from playing around with datasets to obtain what a user would want is going to be pretty large anyway, just from the way notebooks tend to be used. Gaming that would be kind of tough.

With that said, we can easily enough deduplicate queries when we look at telemetry - I imagine if someone is trying to game the telemetry, it would be something like
```python
for _ in range(1e6):
   catalog.search(experiment='my_very_important_experiment', variable='my_very_important_variable')
```

All those calls will leave identical records in the telemetry (besides timestamp & primary key) - so should be straightforward to deduplicate.

rbeucher: We should implement a throttle mechanism to prevent recording identical telemetry calls within a short time window (e.g., a few seconds). This will help mitigate excessive server requests and protect against potential overload.

marc-white:
I've had a prospective data owner ask about the possibility of getting usage statistics/information for the dataset they're proposing to add to the catalog. Does this PR track the usage down to that level?

charles-turner-1:
I would need to double check, but we'll definitely know how many times a particular datastore has been searched for/opened. Potentially with some more work we can make things more granular - the telemetry package is designed to be pretty general and I think we should extend it in whatever direction we find helps collect useful data.

charles-turner-1:
Marking as draft until we have a privacy policy in place.

charles-turner-1:
@rbeucher with the username changes this is ~ ready to merge. 

I haven't uploaded the access-py-telemetry package to conda yet though - how worried are you about environment conflicts if we add this as a pip dependency for the time being in analysis3?

Can either merge & fix conda install later, or wait until we have a conda version before merging?

rbeucher:
I don't think it would be an issue to add it. I have messed up analysis3-25.03 so I need to rebuild it, let me know when you have a package and I can do that.

src/access_nri_intake/data/__init__.py
charles-turner-1: Me & @tmcadam had a look at how many requests we could handle etc. This morning. 

It turns out that the AST implementation is such that any programmatic looping over calls will only trigger a single telemetry record, and nginx appears to already be queueing requests on top of this, so we look to be good on this front.

TLDR;
```python
for _ in range(1e6):
   catalog.search(experiment='my_very_important_experiment', variable='my_very_important_variable')
```
triggers a single telemetry post, as we've only inserted the search node into the AST once.

```python
   catalog.search(experiment='my_very_important_experiment', variable='my_very_important_variable')
   catalog.search(experiment='my_very_important_experiment', variable='my_very_important_variable')
```

will record two records, as there are now two search nodes in the AST.

charles-turner-1:
Tests are failing as production server auth token isn't on Gadi yet. 

Once that's up, we can rerun, verify, & should be ready to deploy.

tmcadam (approved):
There was a bit of discussion in model release today about having payu use a different token. Not for this PR, but could this token path be configured?

charles-turner-1:
I've set it up to just use a `intake_su` user with a token with permissions just on the intake app, so I figure we'd want to just load a different token that we've stuck elsewhere into the payu config.

I feel like I might have misinterpreted your question though....

charles-turner-1:
I've manually verified this is all working on prod now. Ready to go!

@marc-white I can't actually see what your requested changes were - if you could re-review when you get the chance that would be great!

Screenshot 2025-04-02 at 2 41 52 pm

marc-white:
@charles-turner-1 added my two cents on the updates. I think my previous requests were:

- Wondering about emission of warnings, notice to users etc.
- Ability to do statistics on a per-experiment basis

charles-turner-1:
> @charles-turner-1 added my two cents on the updates. I think my previous requests were:
> 
> * Wondering about emission of warnings, notice to users etc.

- I've currently set the telemetry to be completely silent. No warnings, notices, info, print statements, etc. should escape from the telemetry and be shown in the notebook to the user. This seems to be roughly what was agreed on in [this](https://github.com/ACCESS-NRI/access-nri-intake-catalog/pull/312/files/a13e5b8bd5b80954be9da346ce4057a0781ddbc7#r1931580760) comment thread.
- I've added the relevant info to the docs.

> * Ability to do statistics on a per-experiment basis

- Unfortunately, I think this is going to be difficult to do cleanly - at least with the way that I've currently built the telemetry. Consider the following situation:
```python
expt_1 = "some_str"
expt_2 = "some_other_str"

import intake
catalog = intake.cat.access_nri
esm_ds1, esm_ds2 = catalog[expt1], catalog[expt2]

esm_ds1.search(variable="xyz", ...)
esm_ds2.search(variable="abc", ...)

```

As of right now, the telemetry uses some [wacky AST chicanery](https://github.com/ACCESS-NRI/access-py-telemetry/blob/6fb52525a92579e9387605a648eb35afc572c721/src/access_py_telemetry/ast.py#L78) to resolve that `esm_ds1` and `esm_ds2` are of type `esm_datastore`, and then it sends off telemetry records like:
```json
{
    "timestamp": "2025-04-01T01:07:53.146786Z",
    "name": "unknown",
    "function": "esm_datastore.search",
    "args": [],
    "kwargs": {
        "variable": "xyz"
    },
    "session_id": "df07384b-acca-49e9-949c-dbef13b45562",
    "catalog_version": "v2025-03-04"
},
{
    "timestamp": "2025-04-01T01:07:53.146786Z",
    "name": "unknown",
    "function": "esm_datastore.search",
    "args": [],
    "kwargs": {
        "variable": "abc"
    },
    "session_id": "df07384b-acca-49e9-949c-dbef13b45562",
    "catalog_version": "v2025-03-04"
},
```

Without adding some more complicated state tracking to the AST NodeVistor/CallListener, it's going to be tough to know which is which. For example, the `session_id` field is a singleton, so we know it's unique to that interpreter - which removes the need for disambiguation. 
The `ApiHandler` class in the telemetry module contains a `last_call` attribute, which we could extend to cache the last experiment loaded - but in the code block above, both searches would then point at `expt_2`. 

I agree that it would be good to include per-experiment information, but I'm not quite sure how to do it yet.

My hopes is that people will generally only want to interact with a single experiment per notebook, and as such, we could generally associate each `session_id` with a single call to `DfFileCatalog.__getitem__`, which means that we can treat that `session_id` as a proxy for the experiment we would find in the `args` field on the `DfFileCatalog.__getitem__` call for that `session_id`.

charles-turner-1:
Think this is ready to go now then!

PR #315, opened 2025-01-15
Remove bare get_timeinfo, repoint tests to GenericTimeParser

### LINKED ISSUE #314: [BUG] remove deprecated `get_timeinfo` method from `source/utils.py`
# Describe the bug

I think the existing `get_timeinfo` stand-alone method has now been made deprecated by the `TimeParser` objects (@charles-turner-1 is that correct)? If so, we can remove it.

# To Reproduce

N/A

# Additional context

I don't think we need to do a DeprecationWarning first, I doubt anyone is using it external to the package.

marc-white:
Closes #314 .

The simple PR removes `source.utils.get_timeinfo`, which has been superseded by `GenericTimeParser`. The related tests have been switched to use that `Parser`.


charles-turner-1 (commented):
Looks good to me.

Have you got any idea why codecov is so upset about the changes? It's flagged lines don't really make any sense to me, but I wonder whether we might want to take a closer look just to be sure...

marc-white:
> Have you got any idea why codecov is so upset about the changes? It's flagged lines don't really make any sense to me, but I wonder whether we might want to take a closer look just to be sure...

I'm assuming that, because I've excised a chunk of 100% covered code, any uncovered code is now a larger proportion of the code base, hence the percentage of uncovered code has gone up, hence CodeCov is angry... but I would have thought it was smart enough to be able to realize that all that's happened is some line removal.

marc-white:
That being said, you're right, the highlighting is confusing. For example, L117-135 of `source/catalog/utils.py` is showing both a red exclamation point as being 'uncovered', but there's a green highlight next to that I don't understand... plus, local `pytest-cov` shows those lines as tested.

marc-white:
OK, fixed the issue. I'd missed the module-level private helper functions that now weren't being used, so weren't being tested, so Codecov was flagging that correctly as a test coverage lowering.

charles-turner-1 (approved):
Looks good

charles-turner-1:
> OK, fixed the issue. I'd missed the module-level private helper functions that now weren't being used, so weren't being tested, so Codecov was flagging that correctly as a test coverage lowering.

I came across [this extension](https://marketplace.visualstudio.com/items?itemName=tooltitudeteam.tooltitude-py) the other day - might help with that sort of unused function stuff? It basically adds a usage count to function definitions, makes it super easy to see unused functions when traipsing through files.

PR #317, opened 2025-01-22
Individual expts

### LINKED ISSUE #274: Cataloguing of individual experiments

Follows from [Building intake-esm datastores](https://github.com/payu-org/payu/issues/521#event-15447204663) in the Payu repository.

To minimise friction & get users used to the intake catalogue system, it would be good to provide a utility to generate (or open) a catalogue for an experiment just by pointing to the path that contains the outputs.

Functionality to: 
1. Point to the *already known directory* where an experiment output is saved & automatically build an ESM-Datastore for this experiment, if no catalogue is found there.
2. Open the catalog found there, if there already exists a catalog in the directory.
3. If outputs are moved off scratch, the catalog will break (the `catalog.json` file contains an internal reference to it's location & moving a catalog therefore breaks it). In such cases, it would be nice to rebuild the catalog.
4. Hooks to link this functionality to. I think this would in practice just look like an entrypoint that allows us to call this functionality from a bash script.

This might looks something like the following:

```bash
# Generate a datastore for an experiment which hasn't been catalogued yet
$ generate-esm-datastore --builder AccessOm2Builder --experiment-dir $DIR 

Generating esm-datastore...
Datastore successfully written to $DIR/catalog.json!

# Try to generate a datastore for the experiment which we just catalogued
$ generate-esm-datastore --builder AccessOm2Builder --experiment-dir $DIR 

esm-datastore found in $DIR, verifying datastore integrity...
Datastore integrity verified, aborting build

# Move the experiment datastore & catalogue (ie. off scratch)
$ cp -r $DIR $NEWDIR && cd $NEWDIR
$ generate-esm-datastore --builder AccessOm2Builder --experiment-dir $NEWDIR 

esm-datastore found in $NEWDIR, verifying datastore integrity...
Datastore broken due to path inconsistency, regenerating datastore...
Datastore successfully written to $NEWDIR/catalog.json!
```

From within python, we could then have a convenience function that looks something like:
```python
>>> from access_nri_intake import use_esm_datastore
>>> my_datastore_path = "/scratch/abc/xyz/etc/experiment_dir/"
# Generate a datastore for an experiment which hasn't been catalogued yet
>>> use_esm_datastore(experiment_dir = my_datastore_path)
Generating esm-datastore...
No builder supplied - please supply one of `AccessOm2Builder`,...

>>> esm_ds = use_esm_datastore(experiment_dir = my_datastore_path, builder = AccessOm2Builder)
Generating esm-datastore...
Datastore successfully written to /scratch/abc/xyz/etc/experiment_dir/catalog.json!

>>> esm_ds
$EXPERIMENT_NAME datastore with $X dataset(s) from $Y asset(s):

# Run it again on the same dir:
>>> esm_ds = use_esm_datastore(experiment_dir = my_datastore_path, builder = AccessOm2Builder)
esm-datastore found in /scratch/abc/xyz/etc/experiment_dir/, verifying datastore integrity...
Datastore integrity verified, aborting build

# Move it, run again
!cp -r $DIR $NEWDIR
>>> my_new_datastore_path = "/home/abc/xyz/etc/experiment_dir/"
>>> esm_ds = use_esm_datastore(experiment_dir = my_new_datastore_path, builder = AccessOm2Builder)
esm-datastore found in /home/abc/xyz/etc/experiment_dir/, verifying datastore integrity...
Datastore broken due to path inconsistency, regenerating datastore...
Datastore successfully written to /home/abc/xyz/etc/experiment_dir/catalog.json!

>>> esm_ds
$EXPERIMENT_NAME datastore with $X dataset(s) from $Y asset(s):
```

@chrisb13 @anton-seaice are you able to confirm this is the sort of functionality we're after?

charles-turner-1:
Closes #274.

- Adds a `use-esm-datastore` entrypoint that implements the functionality described in #274.


charles-turner-1:
Functionality here is implemented & mostly working - just need to add tests and fix a couple of edge cases.

charles-turner-1:
The current changes in this branch seem to be breaking tests relating to building catalogues: see 
```bash
short test summary info
FAILED tests/test_cli.py::test_build_repeat_adddata - AssertionError: assert 'gdata/None' == 'gdata/al33'
FAILED tests/test_cli.py::test_build_existing_data_existing_old_cat[v2001-01-01-v2099-01-01] - AssertionError: assert 'gdata/dc19+g...10+gdata/tm70' == 'gdata/al33+g...10+gdata/tm70'
FAILED tests/test_cli.py::test_build_existing_data_existing_old_cat[None-v2099-01-01] - AssertionError: assert 'gdata/dc19+g...10+gdata/tm70' == 'gdata/al33+g...10+gdata/tm70'
FAILED tests/test_cli.py::test_build_existing_data_existing_old_cat[v2001-01-01-None] - AssertionError: assert 'gdata/dc19+g...10+gdata/tm70' == 'gdata/al33+g...10+gdata/tm70'
```

charles-turner-1:
Couple of todo's on this before I'll request a review but all CI now passing.

Whoever does review this - I'm sorry, there's a lot of it.

marc-white:
My OS X is still showing a test failure on `test_entrypoint`:

```term
FAILED tests/test_cli.py::test_entrypoint - assert 32512 == 0
```

marc-white:
> My OS X is still showing a test failure on `test_entrypoint`:
> 
> ```
> FAILED tests/test_cli.py::test_entrypoint - assert 32512 == 0
> ```

Strike my last, it worked once I removed & reinstated my editable install of the package.

tests/test_cli.py
```diff
@@ -1096,3 +1100,65 @@ def test_metadata_template_default_loc():
...
+
+        assert f"Builder {builder} is not a valid builder." in str(excinfo.value)
+
+
+@pytest.mark.parametrize(
+    "expt_dir, cat_dir",
+    [
+        ("/not/a/real/dir", "."),
+        (".", "/not/a/real/dir"),
+    ],
```
marc-white, line 1133: I know no-one is going to legitimately have `/not/a/real/dir` on their machine... but on the off-chance that we encounter a computer science professor who decides to break out tests, is there a more robust/randomized way to get a guaranteed-invalid directory path? (Asking mostly out of curiosity, probably doesn't need to actually be changed.)

charles-turner-1: Good question... would `/dev/null/not/a/real/dir` be guaranteed to break?
The requirements would be
- Presumably an absolute path - a relative one should be straightforward to break/fool.
- Something 'unmountable' - hence starting with /dev/null. I think it's impossible to create anything in here - but I haven't got a clue what would happen if you tried.

charles-turner-1: If we restricted ourselves to windows systems, we could make it quite a bit easier - there are a bunch of [reserved directory names](https://help.interfaceware.com/v6/windows-reserved-file-names)

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+    builder: Builder,
+    experiment_dir: Path,
+    catalog_dir: Path | None = None,
+    builder_kwargs: dict | None = None,
+    open_ds: bool = True,
+    datastore_name: str = "experiment_datastore",
+    description: str | None = None,
+) -> esm_datastore | None:
+    """
+    Handles building datastores for experiments for experiments contained in a
```
marc-white, line 32: One-line short description?

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+
+    Procedure:
+    1. Look for a catalog.json and a matching catalog.csv.gz file in the directory
+    provided. TBC: How robust can we make this? Does it need to be in the root of
+    the experiment dir? These things all need to be worked out.
+    2. If we find a catalog there, then we need to check that it is valid. First,
+    we need to just try to open the catalog. If that doesn't work, we need to rebuild
+    it, for sure. But we also need to check that the catalog has the right number of
+    entries, and that the entries are all valid. This is a bit more tricky, but we
+    use just enumerate the entries and check they're all in the directory, and then
```
marc-white, line 45:
```suggestion
    just enumerate the entries and check they're all in the directory, and then
```

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+        The directory containing the experiment.
+    catalog_dir : Path, optional
+        The directory containing/to write the catalog to, if it differs from the
+        experiment directory. If None, the catalog will be written to the experiment
+        directory.
+    open_ds : bool
+        Whether to open the datastore after building it.
+    builder_kwargs : dict, optional
+        Any additional keyword arguments to pass to the builder if needed.
+
```
marc-white, line 64: Needs documentation for datastore_name, description

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+class MultipleDataStoreError(DataStoreError):
+    pass
+
+
+@dataclass
+class DatastoreInfo:
+    """
+    Datastores have a json file and a csv.gz file. This class is a simple way to
```
marc-white, line 31: Add single-line short description

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+    valid: bool = field(default=True)
+    invalid_ds_cause: str = field(default="unknown issue")
+
+    def __post_init__(self):
+        """
+        Run through a bunch of potential issues with the datastore and valid flag
+        and cause accordingly.
+
+        This fails at the first issue it finds. We should find a more comprehensive/elegant/faster
+        way to deal with it, but that's a problem for another day.
```
marc-white, line 52: Once we phase out Python 3.10, apparently there's an ExceptionGroup that we can raise: https://realpython.com/python311-exception-groups/#group-exceptions-with-exceptiongroup

charles-turner-1: I've never come across that feature before - that looks super handy. 

Did you have using exception groups in mind for the (I think) catalog build issues, where it would fail at the first error & cause the user to repeat ad infinitum until they were all fixed?

marc-white: I had not, I searched for raising multiple Exceptions about five minutes ago when I saw this comment :P

For the catalog build issues, I was simply raising a UserWarning every time an individual experiment failed and then moved on. I just had another quick look, and as much as I'd like to use one, I don't think there's an equivalent group for warnings...

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+        look at ds_json["catalog_file"] and check that this exists, using a fsspec
+        get_mapper. If it doesn't exist, then it prepends the dirname of fsspec.get_mapper().root
+        to the path, which winds up creating a horrendously bundled path, something
+        like '/home/189/ct1163/experiments_274/file:///home/189/ct1163/test_datastore_built_in_homedir.csv.gz
+
+        - The reason we need to be careful is that potentially the .name attribute of the Path object
+        might still match, even though the handles are invalid
+
+        We can match fo this pattern with the reget r'.+/file:///.+$
+        """
```
marc-white, line 97: I don't think I've ever seen a string literal mid-method like this before... obviously it has made it past the pre-commit checks (`black`, `ruff`, `mypy`), but is this common?

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+            ]
+        )
+
+
+def verify_ds_current(
+    ds_info: DatastoreInfo,
+    experiment_files: set[Path],
+) -> bool:
+    """
+    Check that the datastore is current - do we have assets in our directory that
```
marc-white, line 144: Single-line short description

src/access_nri_intake/experiment/utils.py
charles-turner-1: That's a really good question - I genuinely have no idea if doing this is common. 

My understanding is that there is nothing (particularly) special about the 
```python
"""
docstring
"""
```
syntax - the python interpreter by default (and I don't know if the default can even be changed) will ignore any unassigned strings, which I guess is a sort of hack to account for the fact python doesn't have multiline comments like most other languages do?

I often see people using this trick in notebooks, but I'm less sure about whether it's common in functions - and common usage in notebooks is definitely not a guarantee of something being a good habit.. more the opposite.

I ran it through copilot as a question (I have absolutely no idea how to search for this pattern across public codebases) and it seems to think that it's generally a bad idea, so I'll excise it to a separate submethod I think...

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+    experiment_files_str = {str(file) for file in experiment_files}
+
+    if experiment_files_str != manifest_files:
+        warn_str = (
+            "extra files in"
+            if len(experiment_files_str) < len(manifest_files)
+            else "missing files from"
+        )
+        warnings.warn(
+            f"{Fore.YELLOW}Experiment directory and datastore do not match ({warn_str} datastore). Regenerating datastore...{Style.RESET_ALL}",
```
marc-white, line 191: If I'm reading this right, `verify_ds_current` isn't actually regenerating the datastore, but signalling that it needs to be regenerated. I'd recommend changing the end of the message to read "Datastore regeneration required..." or similar.

src/access_nri_intake/experiment/utils.py
```diff
@@ -0,0 +1,326 @@
...
+                ret[kw] = True if arg.lower() == "true" else False
+
+    return ret
+
+
+def validate_args(builder: Builder, builder_kwargs: dict[str, Any]) -> None:
+    """
+    Take a builder and validate the kwargs provided against the builder's signature.
+    This is provided to smooth debugging when wrong kwargs are passed from the command
+    line.
```
marc-white, line 286:
```suggestion
    Take a builder and validate the kwargs provided against the builder's signature.
    
    This is provided to smooth debugging when wrong kwargs are passed from the command
    line.
```

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+
+    matched_pairs: list[tuple[Path, Path]] = []
+    for json_file in json_files:
+        for csv_file in csv_files:
+            if (
+                json_file.stem
+                == csv_file.name.replace(
+                    "".join([suffix for suffix in csv_file.suffixes]), ""
+                )  # THis gnarly statement removes the whole suffix to compaer stems
```
marc-white, line 180:
```suggestion
                )  # THis gnarly statement removes the whole suffix to compare stems
```

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+                and json_file.parent == csv_file.parent
+            ):
+                matched_pairs.append((json_file, csv_file))
+
+    if len(matched_pairs) == 0:
+        return DatastoreInfo("", "", False, "")
+    elif len(matched_pairs) > 1:
+        raise MultipleDataStoreError(
+            f"Multiple datastores found in {experiment_dir}. Please remove duplicates."
+        )
```
marc-white, line 190: ...is it actually illegal to have multiple data stores in the same directory? I would have thought that if the datastores were named differently, and referred to different groups of data within the experiment, that would be valid? But I suppose what we really want is for an entire experiment to be encapsulated in a single datastore.

src/access_nri_intake/cli.py
```diff
@@ -573,3 +574,92 @@ def metadata_template(loc: str | Path | None = None) -> None:
...
+        type=str,
+        help=(
+            "Builder to use to create the esm-datastore."
+            " Builders are defined the source.builders module. Currently available options are:"
+            f" {', '.join(builders.__all__)}."
+            " To build a datastore for a new model, please contact the ACCESS-NRI team."
+        ),
+        required=True,
+        # If we can, it would be nice to eventually relax this and try to automatically
+        # determine the builder if possible.
```
marc-white, line 601: That sounds a bit horrifying... perhaps you could push one file through all the Builders in sequence, see which one doesn't crash, and use that? Otherwise, I would think this would involve an enormous amount of very detailed file parsing.

src/access_nri_intake/experiment/main.py
```diff
@@ -0,0 +1,192 @@
...
+    else:
+        # No datastore found. Build one.
+        print(
+            f"{Fore.GREEN}Generating esm-datastore for {experiment_dir}{Style.RESET_ALL}"
+        )
+
+    scaffold_cmd = "scaffold_catalog_entry" if open_ds else "scaffold-catalog-entry"
+    ds_full_path = str((catalog_dir / f"{datastore_name}.json").absolute())
+
+    if not ds_info.valid:
```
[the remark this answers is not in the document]
charles-turner-1: This seems obvious now you've pointed it out - mark of a great idea.

src/access_nri_intake/experiment/main.py
charles-turner-1: I don't think that it's invalid generally speaking, but what we are after in this instance is a datastore for the data in a directory (and it's subdirectories), so that we can pass a path and open it - the only required arguments for `use_datastore` are a builder and a path: 
```python
def use_datastore(
    builder: Builder,
    experiment_dir: Path,
    catalog_dir: Path | None = None,
    builder_kwargs: dict | None = None,
    open_ds: bool = True,
    datastore_name: str = "experiment_datastore",
    description: str | None = None,
) -> esm_datastore | None:
    """
```

Looking at it now, builder should also be optional - in the case that we just want to open the datastore.

Ignoring that for now, multiple datastores in a directory will make using a datastore just by passing the path ambiguous - hence the error.

src/access_nri_intake/cli.py
charles-turner-1: @anton-seaice suggested that the payu run/metadata files from the new(er)(est) versions of payu will provide model info. These aren't actually constrained and can be anything according to @dougiesquire, but it should be possible to provide some sort of many to one mappings to identify the model, eg:
```json
{
    ...
    "model" : "om2",
    ...
}
```
```json
{
    ...
    "model" : "access-om2",
    ...
}
```
```json
{
    ...
    "model" : "ACCESS-OM2",
    ...
}
```
```json
{
    ...
    "model" : "AccessOm2",
    ...
}
```

and then we have something like
```python
OM2_MODEL_LABELS = {"om2", "access-om2""ACCESS-OM2","AccessOm2",}

if payu_metadata.get('model',None) in OM2_MODEL_LABELS:
     do_something()
```

NB. I'm against trying to do anything regex-y/ stringly typed here - I think it'll be more hassle than its worth - eg expressions of the type.
```python
   if re.match("(?i)(access)?.{0,1}om2":
      do_something()
```
 will probably cause more problems than they solve in such instances, IMO.

anton-seaice: I think just using the values for the model field which are in the released configs is a good start

Which seem to follow the normal naming convention for models (e.g. `access-esm1.5`, `access-om2`, `access-om3` ) although if its filled automatically then its presumably the LHS of [this dict](https://github.com/payu-org/payu/blob/083a22b25d5940d9b6ec4ddb51708b405f0d03ed/payu/models/__init__.py#L23) in payu

anton-seaice: @jo-basevi : Just confirming, if a `metadata.yaml` is not provided to payu, then the auto generated one will have a `model` field which is the same as the `model` field specified in `config.yaml` ?

jo-basevi: Yes, if a new experiment is created, it will auto-generate the model field to be a capitalised version of the one in `config.yaml`. There is an `config.yaml` option to override the model name in the generated metadata but I'm unsure it's actually used in existing configs, e.g 
```yaml
metadata:
  model: ACCESS-ESM1.5
```

marc-white: We still need to be able to cover the old runs though, I think... perhaps we only require the `Builder` if its not specified in the metadata, and throw an error if you try to make the system guess?

anton-seaice: Yeah - using the standardish names covers 90% of cases. If it can't figure it out easily, return an error.

charles-turner-1: cool, might be a bit more straightforward than I'd hoped then. 

The dict @anton-seaice linked 
```python
index = {
    'access':     Access,
    'access-om2': AccessOm2,
    'access-om3': AccessOm3,
    'cice':       Cice,
    'cice5':      Cice5,
    'gold':       Gold,
    'matm':       Matm,
    'yatm':       Yatm,
    'mitgcm':     Mitgcm,
    'mom':        Mom,
    'nemo':       Nemo,
    'oasis':      Oasis,
    'test':       Test,
    'um':         UnifiedModel,
    'ww3':        WW3,
    'mom6':       Mom6,
    'qgcm':       Qgcm,
    'cable':      Cable,
    'staged_cable':   StagedCable,

    # Default
    'default':    Model,
    'model':      Model,
}
```
contains a bunch of other models we don't have builders from. Eventually we'll want to add builders for these I suppose - have we got any sense of priority for adding builders for these other models?

Perhaps we'll be able to work it out from the Payu telemetry - hah

anton-seaice: Mostly the others are unimportant I think.

making sure access-esm1.6 still works with the current builder, and the models using the newer UMs, e.g. access-am3/access-ram/access-esm3 would seem to be the priorities

charles-turner-1: For the sake of not making reviewing this PR *even more* of a nightmare, I'd suggest we start with requiring users to specify a builder, and add auto-determining builders later.

tests/test_cli.py
marc-white: The other realization I just had - wouldn't we want to separately test giving a bad `expt_dir` and a bad `cat_dir` separately? At the moment, I think this test is only checking if `use_esm_datastore` gives the expected error for the first one of those two that it checks.

marc-white: (Back to the 'assured fake directory' question, you could also just mock whichever function seeks out the directory to safely achieve the same result.)

tests/test_experiment.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+@pytest.mark.parametrize(
+    "json_name, csv_name, validity, invalid_ds_cause",
+    [
+        (
+            "malformed/missing_attribute.json",
+            "malformed/missing_attribute.csv",
+            False,
+            "columns specified in JSON do not match csv.gz file",
```
marc-white, line 35: Much like the colours, should these error messages be abstracted away into variables so we only maintain them in one place?

tests/test_experiment.py
```diff
@@ -0,0 +1,464 @@
...
+    shutil.copytree(test_data / "esm_datastore" / "local_paths", tmpdir / "local_paths")
+
+    experiment_files = [
+        str(file)
+        for file in Path(tmpdir / "local_paths" / "nc_files" / "barpa-py18")
+        .resolve()
+        .glob("*.nc")
+    ]
+
+    # Mock the builder instance to haev a df.path.tolist() method that returns
```
marc-white, line 176:
```suggestion
    # Mock the builder instance to have a df.path.tolist() method that returns
```

tests/test_experiment.py
```diff
@@ -0,0 +1,464 @@
...
+    experiment_files = [
+        str(file)
+        for file in Path(tmpdir / "local_paths" / "nc_files" / "cmip6-oi10")
+        .resolve()
+        .glob("*.nc")
+    ]
+
+    ds_dir = tmpdir / "local_paths"
+
+    # Mock the builder instance to haev a df.path.tolist() method that returns
```
marc-white, line 215:
```suggestion
    # Mock the builder instance to have a df.path.tolist() method that returns
```

tests/test_experiment.py
charles-turner-1: Yeah, I think that's probably wise... I'll stick them all in an enum.

tests/test_cli.py
marc-white: See: https://stackoverflow.com/questions/57195752/guaranteed-unit-test-fixture-for-non-existent-file . Looks like a UUID might be able to create a pretty-much-guaranteed-not-to-exist filename, although mocking is likely safer to ensure the 'right' error is thrown.

charles-turner-1: > The other realization I just had - wouldn't we want to separately test giving a bad `expt_dir` and a bad `cat_dir` separately? At the moment, I think this test is only checking if `use_esm_datastore` gives the expected error for the first one of those two that it checks.

Maybe I'm misreading your comment, but the first parametrisation gives a valid catalog dir but an invalid experiment dir, and the second vice versa, so I think it is testing both? I'm pretty sure codecov would have caught it if it were only testing one.

src/access_nri_intake/cli.py
marc-white: Concur, get it working first, quality-of-life improvements after.

tests/test_cli.py
marc-white: Oh so it does, I misread that.

marc-white (approved):
I think this is good to go now!

PR #323, opened 2025-02-03
Update meta.yaml

### LINKED ISSUE #322: [BUG] Conda `meta.yml` needs updating
# Describe the bug

See [.conda/environment.yml](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/c476b478d1fd022ba59586325f0667bc558ffab2/.conda/meta.yaml):
in particular:
```yml
    - python >=3.9
 ...
    - intake==0.7.0
```
need updating, or we will wind up with the wrong dependencies pinned (I assume - I've only ever handled Python packaging through PyPI previously, but I'd be very surprised if it's *that* different).

charles-turner-1:
Closes #322.

Updates `meta.yaml` to bump the python version & make sure we don't pin to intake==0.7.0


marc-white (approved):
All good.

PR #327, opened 2025-02-06
326/195 documentation

charles-turner-1:
Closes #326 and addresses most of #195.

- Show users how to use `build-esm-datastore` from command line or in a notebook.
- Add links to the COSIMA tutorial notebook which has some info on time filtering as a stopgap measure on that - GADI is playing up for me. I'll open a new issue to clean that up and move it into this repo. 
- Info about regex, links to core python docs and intake-esm.
-  Minor type checking issue (fell over a Path/str problem myself, users are likely to do the same)

PR #329, opened 2025-02-07
328 datastore names

### LINKED ISSUE #328: Allow for custom ESM-datastore names in `build-esm-datastore`

`use_datastore` allows the user to configure their datastore name, `build-esm-datastore` (the CLI wrapper) does not. 

This makes it difficult if people want to shove all their individual experiment catalogues in the same location, and so we need to expose that config option to the command line interface.

See also  https://forum.access-hive.org.au/t/making-intake-datastore-for-panantarctic/4117/17

charles-turner-1:
Closes #328 .

- I also rewrote the parse_kwargs functionality to use ast.literal_eval instead of some very custom logic. Potentially this could be cleaned up further - I suspect argparse might have something in it that negates the need for it.
- [x] tweak the docs in #327 to reflect this change.

### DIFF
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -632,11 +632,33 @@ def use_esm_datastore(argv: Sequence[str] | None = None) -> int:
         ),
     )
 
+    parser.add_argument(
+        "--datastore-name",
+        type=str,
+        help=(
+            "Name of the datastore to use. If not provided, this will default to"
+            " 'experiment_datastore'."
+        ),
+        default="experiment_datastore",
+    )
+
+    parser.add_argument(
+        "--description",
+        type=str,
+        help=(
+            "Description of the datastore. If not provided, a default description will be used:"
+            " 'esm_datastore for the model output in {--expt-dir}'"
+        ),
+        default=None,
+    )
+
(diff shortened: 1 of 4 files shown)


src/access_nri_intake/cli.py
```diff
@@ -632,11 +632,33 @@ def use_esm_datastore(argv: Sequence[str] | None = None) -> int:
...
+
+    parser.add_argument(
+        "--description",
+        type=str,
+        help=(
+            "Description of the datastore. If not provided, a default description will be used:"
+            " 'esm_datastore for the model output in {--expt-dir}'"
+        ),
+        default=None,
+    )
```
marc-white, line 653: Are the new options being tested in the unit tests?

charles-turner-1: Good question, I (carelessly) assumed that codecov would have complained if they weren't. I'll double check the presumably existing coverage is meaningful.

charles-turner-1: Turns out codecov is not as comprehensive as I'd assumed - cheers for catching that!

marc-white (approved):
OK, think we're good to go!

PR #338, opened 2025-02-14
Conda Deployment issues workaround

charles-turner-1:
For more info, see [here](https://github.com/uibcdf/action-build-and-upload-conda-packages/issues/29#issuecomment-2649951251) and [here](https://github.com/ACCESS-NRI/um2nc-standalone/blob/b44e549ec1ba729ffb525bf4fc80f55fc3e72107/.github/workflows/CD.yml#L49-L63)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -63,7 +63,6 @@ jobs:
             channels:
                 - accessnri
                 - conda-forge
-                - coecms
                 - nodefaults
             EOF
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -50,6 +50,22 @@ jobs:
           auto-update-conda: false
           auto-activate-base: false
           show-channel-urls: true
+          
+      - name: Enforce .tar.bz2 packages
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -50,6 +50,22 @@ jobs:
...
+        shell: bash
+        run: |
+            cat > ~/.condarc << EOF
+            conda-build:
+                pkg_format: .tar.bz2
+            channels:
+                - accessnri
+                - conda-forge
+                - coecms
+                - nodefaults
```
dougiesquire, line 67: Why do we need these? We don't have them in https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/main/.conda/environment.yml

dougiesquire: Possibly worth trying to keep `nodefaults`. I'm sure everything we need is on `conda-forge`. If it works, we could/should also change https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/main/.conda/environment.yml

charles-turner-1: Good call - just copypasta'd the thing with no real thought.

dougiesquire (approved):
LGTM

PR #339, opened 2025-02-16
Reinstate missing `metadata.yaml` template to docs (and other doc tweaks)

marc-white:
The main purpose of this PR is to re-instate the missing `metadata.yaml` template example to `management/building.rst`. It appears to have been dropped from the repo at some point in the past.

While doing this, I also identified that the `sphinx-panels` extension used to create the clickable 'cards' on the documentation home page is something of a dead letter. I've replaced it with the more modern `sphinx-design` package. This has changed the appearance of those cards, but I couldn't work out a quick and easy way of re-styling them.


charles-turner-1 (commented):
Read the docs build failed with this error - I tink you need to add `sphinx_design` to the build requirements [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/6a6d5bc2cd4fa56623f1e9832d44c9faa0b5060d/docs/environment-doc.yaml).

```
Could not import extension sphinx_design (exception: No module named 'sphinx_design')
```

Otherwise, all looks very sensible to me.

Issue #345, opened 2025-02-18
[BUG] Unable to open datastore using default intage catalog version


mmr0:
I'm not able to open a datastore using the default version of the intake catalog. The following code throws a bunch of errors:
```
import intake
catalog = intake.cat.access_nri
om_datastore = catalog["01deg_jra55v13_ryf9091"]
```
while the following works:

```
import intake
catalog = intake.cat.access_nri(version="v2024-12-10")
om_datastore = catalog["01deg_jra55v13_ryf9091"]
```

I wasn't sure how to check what the default version is - perhaps the latest? Apologies is this is missing any important context. These are the modules loaded into my ARE Jupyter session:
```
Singularity> module list
Currently Loaded Modulefiles:
 1) singularity   2) conda/analysis3-24.12(access-med:analysis3)   3) pbs 
```

charles-turner-1:
Hi Madi,

Can you try loading conda/analysis3-25.02 instead of conda/analysis3-24.12?

We made a lot of internal restructuring changes during the last release and I think the updated environment should solve the issue.

mmr0:
Yep can confirm it works with conda/analysis3-25.02 loaded

charles-turner-1:
Awesome, that's good to hear.

@marc-white maybe returning to passing a path, as you suggested in #335 is going to be the right way to go about things - I stupidly didn't consider the issue of legacy environments, which we really ought to support.

marc-white:
@mmr0 glad you got your issue sorted!

@charles-turner-1 is the support of the older environments strictly necessary? I.e., are science results going to be dependent on the environment being used? Or can we just force everyone onto the newer environment?

marc-white:
(Also, we can set things up so either a string or a Path works, if we do want backwards-compatibility.)

charles-turner-1:
> (Also, we can set things up so either a string or a Path works, if we do want backwards-compatibility.)

Yup, this is my intention. I think just finding the exact location where a Path is being passed to `intake-dataframe-catalog` and replacing that with `str(path_obj)` should be a generic fix.

PR #346, opened 2025-02-18
Set include_patterns/exclude_patterns from configuration YAMLs

### LINKED ISSUE #343: Ability to define experiment/builder `include_patterns` and/or `exclude_patterns` via `metadata.yaml`
We've received a data request (#325) that would like us to add a specific subset of files within the experiment directory to the catalog.

At the moment, the `include_patterns` and `exclude_patterns` that `Builder`s use to (not) select files are hardcoded in the `__init__` of each model-specific `Builder`, despite the parent `BaseBuilder` offering these as keyword arguments.

What I'm wondering is if it's worth expanding the use of the experiment `metadata.yaml` to optionally define `include_patterns` and/or `exclude_patterns` in an experiment-specific way? We can always keep the current hard-coded values as the sensible fallbacks for the majority of cases where we don't need the additional flexibility.

What do we think (particularly @dougiesquire and @charles-turner-1 )?

(P.S. I've not worked out if any of the existing `Builder`s are actually relevant to #325, so I might have to create a custom `Builder` anyway - regardless, I can still see this being a possible future issue.)

marc-white:
Closes #343 .

This PR updates all existing builders in `main` to allow the builder `include_patterns` and `exclude_patterns` to be specified from the configuration YAML file on a per-experiment basis. The previously-used defaults are still in place if one/both of those values are not defined in the configuration YAML.

Tests included.

### DIFF
--- a/src/access_nri_intake/source/builders.py
+++ b/src/access_nri_intake/source/builders.py
@@ -366,7 +366,7 @@ class AccessOm2Builder(BaseBuilder):
         r"^ocean.*[^\d]_(\d{2})$",  # A few wierd files in ACCESS-OM2 01deg_jra55v13_ryf9091
     ]
 
-    def __init__(self, path):
+    def __init__(self, path, **kwargs):
         """
         Initialise a AccessOm2Builder
 
@@ -379,8 +379,8 @@ class AccessOm2Builder(BaseBuilder):
         kwargs = dict(
             path=path,
             depth=3,
-            exclude_patterns=["*restart*", "*o2i.nc"],
-            include_patterns=["*.nc"],
+            exclude_patterns=kwargs.get("exclude_patterns") or ["*restart*", "*o2i.nc"],
+            include_patterns=kwargs.get("include_patterns") or ["*.nc"],
             data_format="netcdf",
             groupby_attrs=["file_id", "frequency"],
             aggregations=[
@@ -425,7 +425,7 @@ class AccessOm3Builder(BaseBuilder):
         rf"[^\.]*\.{PATTERNS_HELPERS['om3_components']}\..*?({PATTERNS_HELPERS['ymds']}|{PATTERNS_HELPERS['ymd']}|{PATTERNS_HELPERS['ym']}|{PATTERNS_HELPERS['y']})(?:$|{PATTERNS_HELPERS['not_multi_digit']})",  # ACCESS-OM3
     ]
 
-    def __init__(self, path):
+    def __init__(self, path, **kwargs):
         """
         Initialise a AccessOm3Builder
 
@@ -438,14 +438,15 @@ class AccessOm3Builder(BaseBuilder):
         kwargs = dict(
             path=path,
             depth=2,
-            exclude_patterns=[
+            exclude_patterns=kwargs.get("exclude_patterns")
+            or [
                 "*restart*",
                 "*MOM_IC.nc",
                 "*ocean_geometry.nc",
                 "*ocean.stats.nc",
                 "*Vertical_coordinate.nc",
             ],
-            include_patterns=["*.nc"],
+            include_patterns=kwargs.get("include_patterns") or ["*.nc"],
             data_format="netcdf",
             groupby_attrs=["file_id", "frequency"],
             aggregations=[
@@ -497,7 +498,7 @@ class Mom6Builder(BaseBuilder):
     ]
     TIME_PARSER = GfdlTimeParser
 
-    def __init__(self, path):
+    def __init__(self, path, **kwargs):
         """
         Initialise a Mom6Builder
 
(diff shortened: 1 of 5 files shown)


src/access_nri_intake/source/builders.py
```diff
@@ -379,8 +379,8 @@ def __init__(self, path):
         kwargs = dict(
             path=path,
             depth=3,
-            exclude_patterns=["*restart*", "*o2i.nc"],
-            include_patterns=["*.nc"],
+            exclude_patterns=kwargs.get("exclude_patterns") or ["*restart*", "*o2i.nc"],
```
charles-turner-1, line 382: This could be simplified like so:
```diff
- exclude_patterns=kwargs.get("exclude_patterns") or ["*restart*", "*o2i.nc"],
+ exclude_patterns=kwargs.get("exclude_patterns",["*restart*", "*o2i.nc"),
```

src/access_nri_intake/source/builders.py
```diff
@@ -379,8 +379,8 @@ def __init__(self, path):
...
+            include_patterns=kwargs.get("include_patterns") or ["*.nc"],
```
charles-turner-1, line 383: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -438,14 +438,15 @@ def __init__(self, path):
         kwargs = dict(
             path=path,
             depth=2,
-            exclude_patterns=[
+            exclude_patterns=kwargs.get("exclude_patterns")
```
charles-turner-1, line 441: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -438,14 +438,15 @@ def __init__(self, path):
...
+            or [
                 "*restart*",
                 "*MOM_IC.nc",
                 "*ocean_geometry.nc",
                 "*ocean.stats.nc",
                 "*Vertical_coordinate.nc",
             ],
-            include_patterns=["*.nc"],
+            include_patterns=kwargs.get("include_patterns") or ["*.nc"],
```
charles-turner-1, line 449: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -510,15 +511,16 @@ def __init__(self, path):
         kwargs = dict(
             path=path,
             depth=1,
-            exclude_patterns=[
+            exclude_patterns=kwargs.get("exclude_patterns")
```
charles-turner-1, line 514: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -510,15 +511,16 @@ def __init__(self, path):
...
+            or [
                 "*restart*",
                 "*MOM_IC.nc",
                 "*sea_ice_geometry.nc",
                 "*ocean_geometry.nc",
                 "*ocean.stats.nc",
                 "*Vertical_coordinate.nc",
             ],
-            include_patterns=["*.nc"],
+            include_patterns=kwargs.get("include_patterns") or ["*.nc"],
```
charles-turner-1, line 523: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -579,8 +581,8 @@ def __init__(self, path, ensemble: bool):
         kwargs = dict(
             path=path,
             depth=3,
-            exclude_patterns=["*restart*"],
-            include_patterns=["*.nc*"],
+            exclude_patterns=kwargs.get("exclude_patterns") or ["*restart*"],
```
charles-turner-1, line 584: as above simplification

src/access_nri_intake/source/builders.py
```diff
@@ -579,8 +581,8 @@ def __init__(self, path, ensemble: bool):
...
+            include_patterns=kwargs.get("include_patterns") or ["*.nc*"],
```
charles-turner-1, line 585: as above simplification

marc-white:
@charles-turner-1 the 3.13 test only is giving the YAML `ConstructorError` again. I can reproduce the error on my local machine using a new Python 3.13 environment with all the up-to-date dependencies. Any ideas?

charles-turner-1:
Very weird - I'll take a closer look

charles-turner-1 (commented):
Haven't dug into the yaml constructor error yet - bear with

charles-turner-1:
Okay, there were *a lot* of changes to pathlib with Python3.13: see [release notes](https://docs.python.org/3/whatsnew/changelog.html#changelog).

I'm going to assume that something is subtly broken about the Path constructor, so I'll track this down and fix it. Shouldn't take too long, touch wood.

charles-turner-1:
Yeah, looks like an internal structure change in pathlib, from `pathlib.PosixPath` to `pathlib._local.PosixPath`.

charles-turner-1 (approved):
Now had a good read through the tests too and I think this is good to ship.

I think once it's in main, we ought to build a new catalog against it, as I've tacked the string fix into this PR & so it addresses back-compatibility (ie. #345 ) too.

charles-turner-1:
@marc-white I can't request a review from you since you're the author on the PR, but can you cast an eye over the change I made & check it works locally for you quickly before merging?

marc-white:
Looks good to me, I'll merge.

Issue #344, opened 2025-02-18
[BUG] `use_datastore` not robust for invalid assets


anton-seaice:
# Describe the bug

When I run `use_datastore` on experiment output which contains invalid assets, the datastore is always regenerated, even when there are no changes to the experiment output.

The expected behaviour is that it finds the existing datastore and does not regenerate it.

In general, unmodified configurations won't include invalid assets, so I don't think its a priority to address this.

# To Reproduce

Run 

```
use_datastore(
    '/g/data/tm70/as2285/payu/MOM6-CICE6/archive/',
    builder=builders.AccessOm3Builder
)
```

# Additional context

This is the output:

```
Datastore found in /g/data/tm70/as2285/payu/MOM6-CICE6/archive, verifying datastore integrity...
Parsing experiment dir...

[/g/data/xp65/public/apps/med_conda/envs/analysis3-25.02/lib/python3.11/site-packages/access_nri_intake/experiment/main.py:97](https://are.nci.org.au/g/data/xp65/public/apps/med_conda/envs/analysis3-25.02/lib/python3.11/site-packages/access_nri_intake/experiment/main.py#line=96): DataStoreWarning: Experiment directory and datastore do not match (missing files from datastore). Datastore regeneration required...
  ds_info.valid = verify_ds_current(ds_info, found_experiment_files)

Building esm-datastore...
Sucessfully built esm-datastore!
Saving esm-datastore to /g/data/tm70/as2285/payu/MOM6-CICE6/archive
Successfully wrote ESM catalog json file to: file:///g/data/tm70/as2285/payu/MOM6-CICE6/archive/experiment_datastore.json
Hashing catalog to prevent unnecessary rebuilds.
This may take some time...

[/g/data/xp65/public/apps/med_conda/envs/analysis3-25.02/lib/python3.11/site-packages/access_nri_intake/source/builders.py:200](https://are.nci.org.au/g/data/xp65/public/apps/med_conda/envs/analysis3-25.02/lib/python3.11/site-packages/access_nri_intake/source/builders.py#line=199): UserWarning: Unable to parse 3002 assets. A list of these assets can be found in `.invalid_assets` attribute.
  self.get_assets().validate_parser().parse().clean_dataframe()
[...]

PR #347, opened 2025-02-19
Add expanduser() to paths in use_datastore and hash_catalog

charles-turner-1:
Ran into some issues with homedirs - ie. `~/...`.


charles-turner-1:
Not really sure how to add meaningful tests for these changes without cluttering up homedirs, and in particular risking directory name collisions.

However, everything is covered, & I've tested this manually in an ARE instance - all working correctly.

charles-turner-1:
Actually, I think this would be amenable to end-to-end testing - uuid generated temporary directory we can put in the users homedir, and we use a fixture to clean it up.

Potentially could address #344 with an e2e test in the same hit.

marc-white (approved):
Seems pretty straightforward & handy.

PR #349, opened 2025-02-24
Update AccessCm2Builder name patterns & CLI for Cm2/ESM1-5

charles-turner-1:
- Add a naming pattern to match `ocean_scalar_YYYY_MM.nc`: see [forum post](https://forum.access-hive.org.au/t/access-nri-intake-catalogue-for-cm2-025-simulation/4196/15)
- Fix passing of kwargs to `build-esm-datastore` - unit tests were not correctly capturing behaviour.

Question for @marc-white: I've added a comment in `src/access_nri_intake/experiment/utils.py` about handling kwargs being passed through - I think that the changes in #346 will mean that we might want to expose some extra options to kwargs? Right now the parser fixes typing on the ensemble kwarg, but I think we might want to try to make this function a bit more general & powerful, handling kwargs passed in based on what the builder expects?


src/access_nri_intake/experiment/utils.py
```diff
@@ -293,27 +293,29 @@ def find_experiment_files(
...
+                if not isinstance(arg, bool):
                     raise ValueError
             except (ValueError, SyntaxError):
                 raise TypeError(f"Ensemble kwarg must be a boolean, not {arg}.")
 
-    return ret
+    # Do we use some sort of pattern matching in here to allow for passing through
+    # other kwargs to the builder? This will have changed with #346
+
+    return kw, arg
```
marc-white, line 318: Doesn't the structure of this change mean that we're only ever going to parse one kwarg off the command-line, instead of an arbitrary number of them? Alternatively, the name needs to be altered to be `parse_kwarg` (singular).

src/access_nri_intake/cli.py
```diff
@@ -666,7 +666,9 @@ def use_esm_datastore(argv: Sequence[str] | None = None) -> int:
     builder = args.builder
     experiment_dir = Path(args.expt_dir)
     catalog_dir = Path(args.cat_dir) if args.cat_dir else experiment_dir
-    builder_kwargs = args.builder_kwargs or {}
+    builder_kwargs = (
+        {k: v for k, v in args.builder_kwargs} if args.builder_kwargs else {}
+    )
```
marc-white, line 671: Based on the updated docstring of `experiment.utils.parse_kwargs`, shouldn't this line invoke `parse_kwargs` to do the tuple-to-dict conversion?

charles-turner-1: My understanding is that `parse_kwargs` was invoked by `parser.parse_args(argv)`, which is where we get the list of tuples - rather than just a list of strings. I might be misunderstanding your question though?

marc-white: I think the issue is I can't find where the change in output type of `parse_kwargs` from `dict` to `tuple[str, Any]` has been handled...

src/access_nri_intake/experiment/utils.py
marc-white: I'm not sure that's done it... does L307 `for item in kwargs.split()` need removing?

marc-white (commented):
As for expanding the usability of (the now) `parse_kwarg`: The main point of the function seems to be casting the kwarg value into the correct tyoe, so you could probably just define a variable that maps kwarg name to expected type to make it more generic (I can't think of how to make it so generic that it could magically predict the desired type).

src/access_nri_intake/experiment/utils.py
charles-turner-1: Ahh bugger, of course - no whitespace in those strings, so that line is going to be completely redundant. 

Good catch.

src/access_nri_intake/cli.py
charles-turner-1: Ahh, gotcha. 

I think this is just because the syntax looks similar: 
```python
 builder_kwargs = (
        {k: v for k, v in args.builder_kwargs} if args.builder_kwargs else {}
    )
```
looks an awful lot like iterating over a dictionary - in particular here.
```python
{k: v for k, v in args.builder_kwargs}
```
This is just iterating over a list of two-tuples though - for a dictionary it'd be
```python
{k: v for k, v in args.builder_kwargs.items()}
```
Is that what you meant?

charles-turner-1:
> As for expanding the usability of (the now) `parse_kwarg`: The main point of the function seems to be casting the kwarg value into the correct tyoe, so you could probably just define a variable that maps kwarg name to expected type to make it more generic (I can't think of how to make it so generic that it could magically predict the desired type).

I think we can just enumerate the ones we expect? So, eg. in 
```python

        kwargs = dict(
            path=path,
            depth=3,
            exclude_patterns=kwargs.get("exclude_patterns") or ["*restart*"],
            include_patterns=kwargs.get("include_patterns") or ["*.nc*"],
            data_format="netcdf",
            groupby_attrs=["file_id", "frequency"],
            aggregations=[
                {
                    "type": "join_existing",
                    "attribute_name": "start_date",
                    "options": {
                        "dim": "time",
                        "combine": "by_coords",
                    },
                },
            ],
        )
```

we can just do something like the following:
```python
    if kw == "ensemble":
        ...
    elif kw = "exclude_patterns":
     # Assuming it was passed as ` --builder_kwargs ensemble=False exclude_patterns=*restart*,*nonsense_data* `
     # Then it should come in as kw, arg = ('exclude_patterns', '*restart*,*nonsense_data*')
        exclude_pattern_list = arg.split(',')
```

Since the kwargs are limit in scope/number, I think just enumerating all the possibilites here would be a sensible solution.

src/access_nri_intake/cli.py
marc-white: Ah, I see now. Although it's not confusion about the syntax; it looked like you were just using the straight `args.builder_kwargs`, i.e., whatever the user had stuck on the command line - I couldn't see where `parse_kwarg` was getting involved at all. However, I've now spotted `type=parse_kwarg` where the argument is added above (L617), which I'm assuming applies that function to the received values for that kwarg.

charles-turner-1: Yup - if this seems like a dumb way of doing things to you & something jumps to mind as a better way of doing it, I'm all ears - I'm not convinced what I've done here is a particularly fantastic way of solving the problem, but I can't think of a smarter way.

charles-turner-1:
Closing this as we're going to revamp how the builders work, so I don't think we're going to want to add these patterns.

PR #354, opened 2025-02-27
Bugfixes

charles-turner-1:
- Fixes a couple of bugs with the `use_datastore`/`build-esm-datastore` functionality which arose when trying to create multiple datastores in a single directory: needless checking hashes, warnings about datastore already existing, etc.
- Fix parsing of builder-kwargs


tests/test_experiment.py
```diff
@@ -466,13 +467,13 @@ def test_validate_args(builder: str, kwargs, fails, err_msg):
         (
             "esnmebel=True",
             False,
-            {},
+            ("esnmebel", "True"),
         ),
     ],
 )
 def test_parse_kwargs(kwargs, fails, expected):
```
marc-white, line 474:
```suggestion
def test_parse_kwarg(kwargs, fails, expected):
```

Just to be ridiculously picky.

marc-white (approved):
Send it.

PR #360, opened 2025-03-04
Fail build if all necessary storage can't be accessed

### LINKED ISSUE #358: Catalog build should fail if user doesn't have all necessary storage permissions

In the most recent beta release catalog builds, @charles-turner-1 has run into the issue of the build being incomplete because he inadvertently didn't have the necessary NCI group membership to access new data. 

The catalog build process should fail loudly very close to the start of the process if the user doesn't have all necessary NCI group permissions to access the data.

We could just add additional instructions to the build documentation for the build user to check their group memberships, but a backup/automated check should also be done.

# Additional context

From memory, there are some places in the documentation that auto-compute the required project storage flags - could we leverage off that to do the check?

marc-white:
Closes #358 .

This PR adds a new check step to the catalog build process. It will cause the build to fail (early) with a `RuntimeError` if the user running the build doesn't have access to one or more of the necessary project storage areas.

### DIFF
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -487,7 +487,7 @@ def build(argv: Sequence[str] | None = None):
     storage_flags = "+".join(sorted([f"gdata/{proj}" for proj in project if proj]))
 
     # TODO add permissions check here, before any build directories are made
-    # _confirm_project_access(project)
+    _confirm_project_access(project)
 
     # Now that that's all passed, create the physical build location
     try:
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -200,9 +200,10 @@ def _confirm_project_access(projects: set[str]) -> tuple[bool, str]:
     if len(missing_projects) == 0:
         return True, ""
 
-    return False, f"Unable to access projects {', '.join(missing_projects)} - check your group memberships"
-
-
+    return (
+        False,
+        f"Unable to access projects {', '.join(missing_projects)} - check your group memberships",
+    )
 
 
 def _write_catalog_yaml(
@@ -492,9 +493,7 @@ def build(argv: Sequence[str] | None = None):
     # TODO add permissions check here, before any build directories are made
     _valid_permissions, _err_msg = _confirm_project_access(project)
     if not _valid_permissions:
-        raise RuntimeError(
-            _err_msg
-        )
+        raise RuntimeError(_err_msg)
 
     # Now that that's all passed, create the physical build location
     try:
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -855,8 +911,9 @@ def test_build_repeat_altercatalogstruct(test_data, min_vers, max_vers, tmp_path
         ("v2001-01-01", None),
     ],
 )
+@mock.patch("access_nri_intake.cli._confirm_project_access", return_value=(True, ""))
```
charles-turner-1, line 914: Regarding all these
`@mock.patch("access_nri_intake.cli._confirm_project_access", return_value=(True, ""))`
calls, I wonder whether we might be able to use a fixture to mock calls to `_confirm_project_access` by default/ unless otherwise specified - which would clean up the tests a lot, although obfuscate things a touch.

Don't know if you think it's worth doing so?

src/access_nri_intake/cli.py
```diff
@@ -178,9 +178,34 @@ def _get_project(paths: list[str], method: str | None = None):
     else:  # I know this isn't formally necessary, but I find it easier to read
         project |= {_get_project_code(path) for path in paths}
 
+    project = {p for p in project if p is not None}
+
     return project
 
 
+# Confirm access to all necessary projects
```
charles-turner-1, line 186: Move this inside the docstring?

charles-turner-1 (approved):
Couple of minor comments, mostly whether you think we should enable the mock by default using a fixture for `_confirm_project_access`? I'm on the fence about it, so I'll approve & leave it up to you.

tests/test_cli.py
marc-white: Might be worth a shot, so long as it's still explicit/obvious from the test call that it's being mocked. Otherwise, our successors will probably get confused...

src/access_nri_intake/cli.py
marc-white: Actually it's just duplicated there already, we can delete.

tests/test_cli.py
charles-turner-1: Yeah, this was exactly what made me wonder whether it's smart... Maybe we should just leave it, lots of `@mock.patch` decorators is a real smoking gun.

charles-turner-1: Yeah, sounds good to me.

marc-white: I'll have a go, it's something I've not played with much before and wouldn't mind doing it as an exercise.

marc-white: OK, try that...

charles-turner-1 (approved):
Looks good!

PR #361, opened 2025-03-04
New build check tweaks

marc-white:
Made a couple of small tweaks to the new build checking notebook that should improve speed a little, as well as some new checks. @charles-turner-1 feel free to add anything else you discover might be handy.


charles-turner-1:
Looks like the ROMS branch has gotten wrapped up in this? Changes to the notebook all look good to me, but might be worth cherry picking the only last few out and merging directly from main?

marc-white:
I'd forgotten to update main on Gadi before I branched off it, I've fixed that now - the only change should be the notebook.

charles-turner-1:
Not sure... I've always assumed squash & merge cleans up all the extraneous commits, but I think I did something recently which has made me doubt that, which is why I asked.  

Honestly, probably worth worth worrying about too much - I think lets just merge it, it looks from the main branch history like my concerns are baseless.

PR #364, opened 2025-03-06
Documentation updates

### LINKED ISSUE #353: Improve documentation instructions for creating new `metadata.yaml`

The current instructions for creating an experiment `metadata.yaml` are very sparse - they are limited to how to generate the YAML template, and a repeat of what the template looks like.

However, this means a lot of knowledge is assumed:

- Knowledge of correct YAML structure
- Knowledge of which attributes should be strings, which should be lists, which can be either, etc.

The documentation should be expanded to make the above points explicit.

# Additional context

During most data requests I've actioned, I've had to fix the user's `metadata.yaml`, simply because they didn't get given enough information/instruction to do it correctly.

### LINKED ISSUE #362: Update documentation to include non-NCI/Gadi use cases

#357 has pointed out that there are use cases of `access-nri-intake-catalog` beyond Gadi - specifically, the ability to tie the custom builders in with `intake-esm` to produce datastores of experiments outside Australia.

A new section should be added to the documentation, providing instructions for this use case.

Nil.

# Additional context

Nil.

marc-white:
Closes #353 and closes #362 .

This PR adds documentation updates to:

- Explicitly state the use case of `access-nri-intake-catalog` without Gadi access;
- Clearer instructions on creating an experiment's `metadata.yaml` file;
- Minor style & formatting edits.

A small code edit was also added - the `metadata-template` utility function will now automatically add the correct
`schema_version` constant to the template `metadata.yaml` output.


docs/management/building.rst
```diff
@@ -81,16 +81,71 @@ Each source in the catalog must have an associated :code:`metadata.yaml` file th
...
+.. literalinclude:: metadata.yaml
+   :language: yaml
+
+.. warning:: 
+
+   Your experiment UUID must be **unique** to the experiment. Even if you're adding multiple related experiments,
+   each experiment must have a unique UUID.
+
+   There's nothing special about the UUID value - they're simply meant to be randomly-generated values that 
+   are almost guaranteed to be unique. You can get a UUID value easily from any Unix system by running the
```
charles-turner-1, line 101: I wonder if we should add a link in case people get hung up on almost here? [The wikipedia page](https://en.wikipedia.org/wiki/Universally_unique_identifier) is pretty good.

src/access_nri_intake/cli.py
```diff
@@ -601,10 +601,13 @@ def metadata_template(loc: str | Path | None = None) -> None:
 
     template = {}
     for name, descr in EXP_JSONSCHEMA["properties"].items():
-        if name in EXP_JSONSCHEMA["required"]:
-            description = f"<REQUIRED {descr['description']}>"
+        if "const" in descr.keys():
```
charles-turner-1, line 604: I'm not 100% sure I follow what this addition is doing?

charles-turner-1 (commented):
2 minor queries, otherwise looks great!

src/access_nri_intake/cli.py
marc-white: If the schema defines the `name` should be a `const`, it dumps that straight into the output YAML, instead of the description.

So, before, if you ran `metadata-template`, the first line would be:

```yaml
schema: <The version of the schema (string)>
```

If the user leaves that in and doesn't update it to match the current schema version, `metadata-validate` will fail.

With this change, the first line of `metadata-template` output will be:

```yaml
schema: 1-0-3
```

PR #368, opened 2025-03-11
294 enable catalog version reporting

### LINKED ISSUE #294: Improve catalog version reporting

marc-white:
(Re) closes #294 .

This PR adds version reporting for expired catalog YAML files (with the name format `catalog-YYYYMMDD-YYYYMMDD`). These reports are made only when `available_versions` is called with the `pretty=True` kwarg.


marc-white:
The CodeCov failure can probably be ignored - the lines it is complaining about aren't test for the main catalog version read either.

charles-turner-1:
Everything here seems to work as intended, but I'm not quite sure what the intention is? Am I correctly understanding that the `catalog_versions_old.yaml` file contains essentially a match for deprecated catalogues?

I think my confusion is mostly leading from it not seeming to apply (as of yet) to any catalogues that we can obtain with the live catalog via `available_versions()`?

marc-white:
Correct, this doesn't apply to anything we have yet. The idea is that, when you make an update that makes the old catalog YAML incompatible with the new one, you can move the old catalog YAML aside. The system will then report its presence and available versions. The user (who one would presume is a power-user) would then be able to access the old catalogs by dropping the right expired YAML in their home area.

charles-turner-1:
Gotcha - seems sensible. 

I'm just going to take a slightly closer look at codecov - there have been a couple of times recently where I've assumed it's a glitch out only to find I've missed something important.

marc-white:
Oh the Codecov complaint is valid, there are some new lines that aren't tested - however, they're analogous to some existing lines that we also didn't try testing.

charles-turner-1 (approved):
Added a test for deleted invalid catalog files to shut up Codecov.

Looks ready to send!

PR #369, opened 2025-03-12
Translator metadata fixes

charles-turner-1:
Around the 1.0.0 release, I added a decorator to report failing translators, but the implementation was a bit messy - wasn't preserving metadata correctly, circular import workarounds, etc.

This PR cleans that up. Low priority.

### DIFF
--- a/src/access_nri_intake/catalog/utils.py
+++ b/src/access_nri_intake/catalog/utils.py
@@ -1,3 +1,4 @@
+from functools import wraps
 from typing import Callable
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


src/access_nri_intake/catalog/utils.py
```diff
@@ -21,8 +22,33 @@ def tuplify_series(func: Callable) -> Callable:
...
+    Decorator that wraps a function and prints a message if it raises an exception
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        func_name = func.__name__
+        colname = func_name[1:].split("_")[0]
+        # Check that the first argument is a DefaultTranslator (or subclass) instance
+        # This is hacky, but I want the decorators outside the main translators module
+        if not str(type(args[0]).mro()).__contains__("DefaultTranslator"):
```
marc-white, line 44: Does a straight `isinstance` call not work here?

charles-turner-1: It requires an import of the class, which leads to a circular import issue.

Obviously having these decorators all in one file with the translators would solve that, but it seemed cleaner to me having them off in a separate file?

marc-white: That's fine, sounds like a reasonable solution then.

marc-white (approved):

PR #371, opened 2025-03-13
Make ACCESS-NRI Intake Catalog more portable to other computing environments (sort-of)

### LINKED ISSUE #363: Portability
An interesting question was raised by @charles-turner-1 this morning - how portable is `access-nri-intake-catalog`? For example, would another institute be able to pick it up and use it to manage/store their own collection of climate models without having to make structural changes to the code base? (As in, changes to the base-level code, not just adding Builders/Translators are required.)

What jumps out to me at first is:

- I think most of the paths that are set inside the package are variables, so that should be able to be easily modified;
- I think the code that deals with storage flags, storage access etc. is hard-coded to expect `/g/data`-style paths - that would need to change if we wanted to make the package more portable.

marc-white:
Closes #363 .

This PR is my first pass at trying to extract out as much of the Gadi dependency as I can from the main source code. I've also included a quick write-up of what I think would be required to get the code working on a different HPC system.

Doubtless I've missed something, so I'm marking this as a draft for others to look over.


src/access_nri_intake/__init__.py
```diff
@@ -8,4 +8,16 @@
 __version__ = _version.get_versions()["version"]
 
 CATALOG_LOCATION = "/g/data/xp65/public/apps/access-nri-intake-catalog/catalog.yaml"
```
charles-turner-1, line 10: There's a trick we use for managing config files which is common in Django/Web Development which might be relevant/useful here:

```python
import os

CATALOG_LOCATION = os.environ.get("CATALOG_LOCATION", "/g/data/xp65/public/apps/access-nri-intake-catalog/catalog.yaml")
```

Basically, it lets us override the default with an environment variable, if found, falling back to the hardcoded default. It strikes me that it might be useful here, although we'd probably want to prefix all the environment variable names as a namespacing strategy, eg.:
```python
CATALOG_LOCATION = os.environ.get(
    "ACCESS_NRI_INTAKE_CATALOG_LOCATION", "/g/data/xp65/public/apps/access-nri-intake-catalog/catalog.yaml"
)

File: src/access_nri_intake/__init__.py [hunk 2]
```diff
@@ -8,4 +8,16 @@
...
+"""Location for 'live' master catalog YAML."""
+
 USER_CATALOG_LOCATION = str(Path.home() / ".access_nri_intake_catalog/catalog.yaml")
```
charles-turner-1, line 13:
```suggestion
USER_CATALOG_LOCATION = str(Path.home() / ".access_nri_intake_catalog/catalog.yaml")
USER_CATALOG_LOCATION = os.environ.get(
    "ACCESS_NRI_INTAKE_USER_CAT_LOCATION", str(Path.home() / ".access_nri_intake_catalog/catalog.yaml")
)
```

src/access_nri_intake/__init__.py
```diff
@@ -8,4 +8,16 @@
...
+"""Location where user can place a master catalog YAML to override standard 'live' version."""
+
+STORAGE_ROOT = "/g/data"
```
charles-turner-1, line 16:
```suggestion
STORAGE_ROOT = os.environ.get(
    "ACCESS_NRI_INTAKE_STORAGE_ROOT", "/g/data"
)
```

src/access_nri_intake/__init__.py
```diff
@@ -8,4 +8,16 @@
...
+"""Root storage location for catalog experiments"""
+
+STORAGE_FLAG_PATTERN = r"gdata/[a-z]{1,2}[0-9]{1,2}"
```
charles-turner-1, line 19:
```suggestion
STORAGE_FLAG_PATTERN =  = os.environ.get(
    "ACCESS_NRI_INTAKE_STORAGE_FLAG_PATTERN", r"gdata/[a-z]{1,2}[0-9]{1,2}"
)
```

src/access_nri_intake/__init__.py
```diff
@@ -8,4 +8,16 @@
...
+"""Pattern for matching 'storage flags' - related to Gadi file access system"""
+
+STORAGE_LOCATION_REGEX = r"^/g/data/(?P<proj>[a-z]{1,2}[0-9]{1,2})/.*?$"
```
charles-turner-1, line 22:
```suggestion
STORAGE_LOCATION_REGEX = os.environ.get(
    "ACCESS_NRI_INTAKE_STORAGE_LOCATION_REGEX", r"^/g/data/(?P<proj>[a-z]{1,2}[0-9]{1,2})/.*?$"
```

docs/management/external.rst
```diff
@@ -0,0 +1,39 @@
...
+      The build will be aborted if these checks fail. Therefore, if your storage does not
+      use a similar directory structure to Gadi (that is, a group of directories all situated at one root location), you may need
+      to modify or remove these calls to achieve a successful catalog build.
+
+2. The existing YAML files in :code:`config/` refer to the datastores/raw data stored on Gadi. You will need to 
+   remove these, and replace them with similarly-structured YAML files denoting your own data setup. (Note that the 
+   contents of :code:`config/metadata-sources` are archival copies of live experiment :ref:`metadata`; 
+   you will not need to replace these on your system.)
+
+3. The command-line scripts in :code:`bin/` contain PBS commands and file paths specific to Gadi. You will need
```
charles-turner-1, line 38: We should then be able to just update all this to reflect that the defaults can be overridden with environment variables.

charles-turner-1 (commented):
Sorry, somehow never got round to reviewing this.

I think we can use environment variables to make quite a bit more straightforward to port. Otherwise looks good!

PR #374, opened 2025-03-17
Update metadata-template CLI

### LINKED ISSUE #372: [BUG] Stray `metadata.yaml` appears in PWD when running tests
# Describe the bug

While running the `pytest` suite, a `metadata.yaml` appears in the present working directory as a result of one of the tests. The file disappears after the tests complete, although it will remain in place if the test run is aborted for any reason (e.g., quitting after a PDB session).

# Additional context

I'm assuming this is a side effect of testing the `metadata-template` command line function.

marc-white:
Closes #372 .

It turns out the stray `metadata.yaml` came from the `test_entrypoint` test. Because `metadata-template` did not originally use the `argparse` structure, calling `metadata-template --help` did nothing special - it just executed the command with no arguments, and hence wrote a `metadata.yaml` in the CWD with no teardown to remove it. (We didn't notice earlier because the later test that intentionally tested the CWD template write *does* have teardown, so was covering up the issue.)

This PR:
- Adds correct `argparse` usage to `metadata-template`, so `test_entrypoint` now works as expected;
- Updates the `metadata-template` tests to use correct `argparse` formatting;
- Added a per-function pytest fixture to raise a `UserWarning` if a `metadata.yaml` is left hanging in the CWD after each test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,12 +4,12 @@
 from datetime import datetime
 from pathlib import Path
 
-from pytest import fixture
+import pytest
 
 here = Path(__file__).parent
 
 
-@fixture(scope="session")
+@pytest.fixture(scope="session")
 def test_data():
     return Path(here / "data")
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -18,26 +18,37 @@ def metadata_sources():
...
 
+@pytest.fixture(autouse=True, scope="function")
+def check_metadata_cwd():
+
+    # Run test
+    yield
+
+    # Look for metadata.yaml in CWD
+    if (Path.cwd() / "metadata.yaml").is_file():
+        raise UserWarning("Stray metadata.yaml left in CWD!")
```
charles-turner-1, line 49: Would it make more sense to add a 
```
 if (Path.cwd() / "metadata.yaml").is_file():
        (Path.cwd() / "metadata.yaml").unlink()
```
like in the `tests/test_cli.py` file below? We would only fail to remove the file in situations where the warning wasn't triggered (hard exits) anyway - doesn't seem to be any reason to not just remove it to me?

tests/conftest.py
```diff
@@ -18,26 +18,37 @@ def metadata_sources():
...
     yield tmp_path_factory.mktemp("catalog-dir")
 
 
-@fixture(scope="session")
+@pytest.fixture(scope="session")
 def v_num():
     return datetime.now().strftime("v%Y-%m-%d")
 
 
+@pytest.fixture(autouse=True, scope="function")
```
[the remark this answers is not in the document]
marc-white: That's probably a good idea; I was originally looking for all other places that might cough up a `metadata.yaml` by accident.

tests/conftest.py
marc-white: My only concern with that would be that we have no guarantee where the test suite is run from, so potentially we might end up deleting something belonging to the user. I know that's a remote concern, but it's the reason I added a warning, not a delete.

(The second reason is that the warning is a flag to go and fix the test!)

charles-turner-1: Ahh yeah, I hadn't thought of that - that seems sensible to me.

charles-turner-1 (approved):

PR #382, opened 2025-04-09
Fix conda deps

charles-turner-1:
Sorry - screwed up & forgot to add the deps in `.conda/meta.yaml` too. This will have broken the catalog in 25.04 & will need a new release.

Dougie sent me something that could be used to automate syncing these dependencies up, so I'll integrate that before I make the same cockup again.


marc-white (approved):
See no issues here.

PR #386, opened 2025-04-28
385 Synchronise dependencies

charles-turner-1:
Split requirements out into requirements.txt & read that into pyproject.toml/meta.yaml to avoid dependency mismatch issues

More or less follows the approach from [here](https://github.com/payu-org/payu/commit/14bba8c741d6ad7618a8ec81e13098cd0e0376a8) but is a little closer to the way intake-esm structures things.


.conda/meta.yaml
```diff
@@ -1,5 +1,6 @@
 {% set data = load_setup_py_data(setup_file='../setup.py', from_recipe_dir=True) %}
 {% set version = data.get('version') %}
+{% set requirements = load_file('../requirements.txt').splitlines() %}
```
dougiesquire, line 3: I don't see `load_file` mentioned [here](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#loading-data-from-other-files)? I assume you've tested this? Also, does this need `from_recipe_dir=True`?

dougiesquire (commented):
Thanks @charles-turner-1. One probably stupid question below.

.conda/meta.yaml
charles-turner-1: Appears it isn't possible to separate the requirements out of `pyproject.toml` & stick them into a `requirements.txt` here - shame as a single source of truth both files read seems cleaner to me but hey-ho.

Does seem like it could be done with something resembling `requirements.yaml` or `requirements.json`, but since those are (very) nonstandard I'm not gonna do that. I've updated the approach to be the same as what you added to payu.

charles-turner-1: Confirmed changes now build locally.

dougiesquire (approved):
LGTM - thanks @charles-turner-1!

Issue #378, opened 2025-04-01
Rewrite builders to extract relevant info from netCDF files, rather than regex matching on filenames


charles-turner-1:
# Is your feature request related to a problem? Please describe.

The current regex matching approach used to generate file ID's & then separate the experiment into datasets needs to be updated, as we have several datasets for which this approach fails currently in the catalog & we're liable to create more. Additional datasets such as those on #349 will eventually lead to the regex approach becoming a frankensteins monster of string processing as we introduce more layers of workarounds.

# Describe the feature you'd like

When we open up the datasets to build the datastore [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/fe3bf7900ea28451488dfe1e698040adf0cb5558/src/access_nri_intake/source/builders.py#L327), we should pull out the extra info necessary to disambiguate the experiment outputs/file structure into datasets.

dougiesquire:
Related: https://github.com/ACCESS-NRI/access-nri-intake-catalog/issues/67

marc-white:
I've been looking at this today. There's a couple of easy wins we can make in progressing this, and then one major sticking point.

Wins
-----
- I've already removed `timestamp_filename` from the `_NCFileInfo` object. AFAIK, it seems to not be used for anything anymore, so was a dead letter.
- I think I can remove the `filename_frequency` from the file parser fairly easily, and simply leave the frequency as the `STATIC_FREQUENCY` (`fx`) default if the frequency can't be calculated from the file data. I can't remember where, but I think we've discussed this as being a reasonable solution in the past, as it ultimately represents a data issue.

Sticking point
--------------
... is the **realm** information. At the moment, this is being parsed from the filename, or in the case of, say, `ROMSBuilder`, it's being hard-coded to a single value. However, these approaches won't work under this ticket:
- Obviously, we're trying to move away from using filenames for anything data-related;
- Several models have multiple possible realms, so we can't just hard-code a solution.

I've been looking through just the test data contained in the repository. Some models (e.g., `ACCESS-OM2`) seem to contain some useful data in the `X.attrs` dict (where `X` is the `xarray.Dataset` return of `xarray.open_dataset`); we could try parsing that. However, some of the other test data (e.g., `ACCESS-CM2`) seems to contain no explicitly useful information to compute the realm from.

[...]

PR #407, opened 2025-05-21
New build check updates

marc-white:
## Change Summary

This PR makes some slight tweaks to the build script, and relevant updates to documentation.

## Related issue number

Nil.

## Checklist

- [x] Tests pass on CI


src/access_nri_intake/source/builders.py
```diff
@@ -234,67 +234,36 @@ def parser(file):
         raise NotImplementedError
 
     @classmethod
```
charles-turner-1, line 236: I think this is related to #378 rather than this PR, no?

charles-turner-1:
I think a bunch of stuff from the #378 branch has found it's way into this

marc-white:
Yeah I saw that, no clue how that's happened - I'll clean it up shortly.

marc-white:
OK I think I've unstuffed things now.

bin/build_all.sh
```diff
@@ -26,9 +26,11 @@ if [ ! $# -eq 0 ]; then
 fi
 
 module use /g/data/xp65/public/modules
-module load conda/analysis3-25.02 # THIS NEEDS TO BE UPDATED TO THE LATEST VERSION
+module load conda/analysis3-25.05 # THIS NEEDS TO BE UPDATED TO THE LATEST VERSION
 module load openmpi
 
+export PYTHONTRACEMALLOC=1
```
charles-turner-1, line 32: What are we using PYTHONTRACEMALLOC for here?

charles-turner-1 (approved):
Tangential question about what we're using tracemalloc for, but otherwise looks good to me

bin/build_all.sh
marc-white: It passes through the exception that caused any particular experiment to fail to be included.*

*Well, sort of... annoyingly, it doesn't give you the actual exception message, but the line & code that raised the exception.

charles-turner-1: ahh, handy. Good to know, cheers

Issue #398, opened 2025-05-19
[BUG] Catalog build atomicity


charles-turner-1:
# Describe the bug
See also #397 which temporarily fixes this by loosening exception handling.

When importing the catalog via `intake.cat.access_nri`, this code is executed:
```python
try:
    data = intake.open_catalog(get_catalog_fp()).access_nri
    cat_version = data._captured_init_kwargs.get("metadata", {}).get(
        "version", "latest"
    )  # Get the catalog version number and set it to "latest" if it can't be found
except FileNotFoundError:
    warnings.warn(
        "Unable to access a default catalog location. Calling intake.cat.access_nri will not work.",
        RuntimeWarning,
        stacklevel=2,
    )
    data = intake.catalog.Catalog()
    cat_version = -999  # Failed to load a catalog - bad result flag
finally:
    configure_telemetry(["--enable", "--silent"])
    api_handler.add_extra_fields("intake_catalog", {"catalog_version": cat_version})

    api_handler.send_api_request(
        service_name="intake_catalog",
        function_name="intake.cat.access_nri",
        args=[],
        kwargs={"version": cat_version},
    )
```
This will error if any error other than a `FileNotFoundError` is raised during the attempt to open the catalog, further triggering a telemetry error. During catalog builds (in particular I think failed catalog builds), it appears the relevant files are created before being populated. This leads to the error in #397 - where we get a `pandas.errors.EmptyDataError`.

To fix this, we need to ensure that the catalog build is an atomic transaction, which doesn't appear to currently be the case.
# To Reproduce

This is a bit difficult - it requires a failed catalog build as the most recent catalog. The relevant stack trace is shown in #397 .

# Additional context

This could additionally be handled by using a more sophisticated set of `try/catch` statements, eg.
```python

try:
    data = intake.open_catalog(get_catalog_fp()).access_nri
    cat_version = data._captured_init_kwargs.get("metadata", {}).get(
        "version", "latest"
    )  # Get the catalog version number and set it to "latest" if it can't be found
except FileNotFoundError:
    warnings.warn(
        "Unable to access a default catalog location. Calling intake.cat.access_nri will not work.",
        RuntimeWarning,
        stacklevel=2,
    )
    data = intake.catalog.Catalog()
    cat_version = -999  # Failed to load a catalog - bad result flag
except EmptyDataError:
       warnings.warn(
        "Catalog data is empty. Most recent catalog build has failed",
        RuntimeWarning,
[...]

PR #408, opened 2025-05-21
398 atomic build

charles-turner-1:
## Change Summary

- Catalog build all happens in a hidden temporary directory, only move everything out to the final catalog location upon success. Ie. `version = v2025-05-22` builds in `.v2025-05-22`, and is only extracted to `v2025-05-22` after everything has successfully built.
- I haven't added any graceful cleanup due to unexpected failures/stops since I figure a PBS job being killed will stop that in it's tracks. Maybe worth doing so though.

## Related issue number

#398 

## Checklist

- [ ] Tests pass on CI


charles-turner-1:
@marc-white still a long way to go with tests/ etc for this to be ready I think this is ready to take a look at whenever.

src/access_nri_intake/cli.py
```diff
@@ -520,9 +532,50 @@ def build(argv: Sequence[str] | None = None):
...
+        ).write_json(f)
+
+    # Now unhide the directory containing the catalog
+    (Path(build_base_path) / f".{version}").rename(Path(build_base_path) / version)
+
+    # Move the catalog.yaml file to the new location
+    shutil.move(
+        Path(build_base_path) / version / "catalog.yaml",
+        Path(build_base_path) / "catalog.yaml",
+    )
```
marc-white, line 577: Technical question - why use `shutil` here, instead of doing another `Path.rename()`?

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "intake-dataframe-catalog>=0.3.1",
     "intake-esm>=2025.2.3",
     "jsonschema",
+    "polars>=1.24.0",
```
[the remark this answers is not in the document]
charles-turner-1: More or less - I figured since it'll be required through that, I might as well foister my hobby-horse upon everyone here too.

More seriously, it just makes it very straightforward to rename everything in a consistent way - see lines ~555, 565 in `cli.py`. I don't think that can be done with pandas quite so neatly?

src/access_nri_intake/cli.py
charles-turner-1: I didn't find `Path.rename()` when I went looking for that functionality - will update it now

pyproject.toml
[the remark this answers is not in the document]
marc-white: Probably not, and I have zero issue with the foister of hobby-horses :P

charles-turner-1:
@marc-white would you be able to take a look at the remaining failing test for me and tell me the intention? I'm a little confused & I can't get the test to pass (without breaking loads of others)

marc-white:
The test that is failing (`test_build_separation_between_catalog_and_buildbase`) is making sure that the catalog versioning logic ignores any existing catalog 'version folders' that are in the wrong place. This is because the code before the versioning update didn't differentiate between the `build_base_path` and `catalog_path`, so I needed a test to ensure I'd correctly separated the two.

I think you just need to update one of the test paths the test creates with a `"."` prefix, but I'm not entirely sure which one...

charles-turner-1:
Think this is ready for review.

Re. the two unchecked boxes, I'm not quite sure whether we do actually want to add tests for these changes & until we separate out the `_concretize` step I don't think we'll need any documentation changes - but maybe we want to do that in this PR?

src/access_nri_intake/cli.py
```diff
@@ -145,7 +146,15 @@ def _parse_build_directory(
     build_base_path: str | Path, version: str, catalog_file: str
 ) -> tuple[Path, Path, Path]:
     """
-    Build the location for the new catalog
+    Build the location for the new catalog. We put everything in temporary directory
+    for now, which will basically be a map `$DIR/$FNAME` => `$DIR/.$FNAME` where
+    oreakpoint()
```
marc-white, line 151: I think something might have been lost/mangled here?

src/access_nri_intake/cli.py
```diff
@@ -156,9 +165,9 @@ def _parse_build_directory(
     catalog_file : str
         Catalog file name
     """
-    build_base_path = Path(build_base_path).absolute()
-    build_path = Path(build_base_path) / version / "source"
-    metacatalog_path = Path(build_base_path) / version / catalog_file
+    build_base_path = Path(build_base_path)
```
marc-white, line 168: What was the reason for losing the `.absolute()` part of this call? My first impression is that it makes the build dependent on the directory it is run from, but happy to be corrected.

marc-white (approved):
Should be ready to go, pending the last two comments.

src/access_nri_intake/cli.py
charles-turner-1: yeah, think that was an accident.

src/access_nri_intake/cli.py
charles-turner-1: I think this was just an accident - checking now.

charles-turner-1: Yup, just an accident.

PR #424, opened 2025-06-05
Use `pytest-xdist`

charles-turner-1:
## Change Summary

- Add `pytest-xdist` with `-n auto` flag.
- Not sure if this will speed anything up (actually slower on my machine), but it'd be good to see on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Run tests
         shell: bash -l {0}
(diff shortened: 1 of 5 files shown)


charles-turner-1:
- Tests on CI were about 40s, this reduces them to about 30s it seems. 
- Will need some updates to get this to work with coverage by the looks of things. 

I'm gonna mark this as draft & leave it dangling for a rainy day.

charles-turner-1:
Actually, before we merge this, I'll add a pixi task to dump a conda environment.

PR #449, opened 2025-07-18
Fix WOA Builder

### LINKED ISSUE #439: [BUG] WOA Datastore path dependence
# Describe the bug

The WOA builder currently groups together a few datasets, which requires hacks like this:

```python
woa13_10 = cat_subset["WOA-13"].search(path="/g/data/ik11/observations/woa13/10/.*mom10.nc").to_dask(xarray_open_kwargs={"use_cftime": True})
woa13_10_KDS75 = cat_subset["WOA-13"].search(path="/g/data/ik11/observations/woa13/10_KDS75/.*mom10.nc").to_dask(xarray_open_kwargs={"use_cftime": True})
```

I think this is just down to the file naming scheme - we should be able to fix it by changing (I think) the `parse_filename` method on the `WoaBuilder` class to differentiate based on the directory too.

# Additional context

See https://github.com/COSIMA/cosima-recipes/pull/355#issuecomment-3014490750

charles-turner-1:
## Change Summary

- Uses a bit of a hack to work out if we're using a KDS dataset or not. 
- Adds tests

N.B: When we move to grid based `file_id`'s, this will be made redundant & we can remove it, so I'm not too bothered about this being a bit of a janky solution.

## Related issue number

Closes #439 .

Will also allow us to close https://github.com/COSIMA/cosima-recipes/issues/537

## Checklist

- [x] Tests pass on CI


src/access_nri_intake/source/builders.py
```diff
@@ -751,17 +751,30 @@ def __init__(self, path, **kwargs):
...
+    @staticmethod
+    def post_proc_ncinfo(ncinfo_dict: dict) -> dict:
+        """
+        Post process the ncinfo_dict to add anything that looks like `KDS` to the file id
+        """
+        file_id = ncinfo_dict["file_id"]
+        pathparts = Path(ncinfo_dict["path"]).parts
+        for part in pathparts:
+            if "kds" in part.lower():
+                file_id = f"{file_id}_{part}"
```
marc-white, line 772: Should this `break` after reaching the `if` statement? Otherwise you might get multiple `_part` added to the file_id... unless that's the intended behaviour.

marc-white (changes requested):
Need to review L770-772 in `source/builders.py` before merge.

marc-white (approved):

Issue #445, opened 2025-07-10
[BUG] Xarray v2025.07.1 breaking changes


charles-turner-1:
# Describe the bug

Some tests (lots of them) have broken due to changes in (what looks like) stored test data and how it is now constructed on the fly

# To Reproduce

See eg. https://github.com/ACCESS-NRI/access-nri-intake-catalog/actions/runs/16206530294/job/45757983827

charles-turner-1:
It appears that this can't be fixed just by pinning xarray to < 2025.7.0.

Will dig in further when I understand exactly what's happened.

marc-white:
Can confirm tests on my local machine are now failing with `xarray=2024.11.0`

marc-white:
I've managed to get my local tests passing again by setting any hashing reference to `binhash-xxh` back to `binhash`. Is it possible the `binhash-xxh` scheme is coming from another dependency that's not/no longer being installed?

marc-white:
No, I was getting failures out of `test_expirement` complaining that it didn't know what `binhash-xxh` was...

charles-turner-1:
Interesting. Heres a couple of CI run excerpts, and their diff: both with yamanifest 0.3.12

Working (last working CI run on main): 
`yamanifest                          0.3.12                 py_0                     accessnri`

Broken (first failing CI run):
`yamanifest                          0.3.12                 py_0                     accessnri`

Both the same (unsurprising, see the diff):

```diff
#  diff -w env_broken.txt env_working.txt
10a11
> asciitree                           0.3.3                  py_2                     conda-forge
19c20
< aws-c-io                            0.20.1                 h6b9530e_1               conda-forge
---
> aws-c-io                            0.20.1                 hdfce8c9_0               conda-forge
26,31c27,31
< azure-core-cpp                      1.15.0                 h5cfcd09_0               conda-forge
< azure-identity-cpp                  1.11.0                 hb5324b0_1               conda-forge
< azure-storage-blobs-cpp             12.13.0                hf182047_2               conda-forge
< azure-storage-common-cpp            12.10.0                h40e822a_1               conda-forge
< azure-storage-files-datalake-cpp    12.12.0                h141ff2a_2               conda-forge
< beautifulsoup4                      4.13.4                 pyha770c72_0             conda-forge
---
> azure-core-cpp                      1.14.0                 h5cfcd09_0               conda-forge
[...]

PR #432, opened 2025-06-17
Switch to file_id based off file coordinates, not file name

marc-white:
## Change Summary

This PR represents a major change in catalog build process. Now, rather than grouping files into dataset by timestamp-redacted filename, they will be grouped based on the coordinates available within the file itself.

## Related issue number

Closes #378 .

## Checklist

- [x] Unit tests for the changes exist
- [x] Tests pass on CI
- [ ] Documentation reflects the changes where applicable

## Notes

The new file_id is built from the coordinates in the file, alphabetically ordered, e.g., a very long example from the `bz687a` data: `bnds:2.depth:6.lat:144.lat_v:145.lon:192.lon_u:192.model_rho_level_number:85.model_theta_level_number:85.pressure:19.pseudo_level:5.pseudo_level_0:6.pseudo_level_1:17.pseudo_level_2:13`

- I have successfully build a new test catalog using this branch. It is available (read-only for members for NCI group `xp65_w`) at: `/scratch/tm70/mcw120/intake-cat-test/catalog.yaml`
- I have created a short testing Jupyter notebook that is also read-only accessible for `xp65_w` members, at: `/scratch/tm70/mcw120/file_id_change.ipynb`. Feel free to have a play with it. In particular, things I haven't worked out yet are:
   - Have we lost any assets? I've not worked out an efficient way to count those on a per-experiment basis yet.
   - Does having, in general, fewer datasets per experiment blow out our data read-in times?

Issue #419, opened 2025-06-04
[BUG] End-to-end tests broken


charles-turner-1:
# Describe the bug

The atomic catalog build PR (#408) broke these, needs fixing.

PR #450, opened 2025-07-18
Remove unnecessary TimeParser abstraction, back to simpler old functional approach 

### LINKED ISSUE #448: Investigate whether separate MOM6 time parsing logic still necessary

#293 introduced some `TimeParser` classes in order to abstract out the parsing of times from files.  This was necessary because we were getting a bunch of test failures related to integer under/overflow. 

#446 replaces the data in these tests which was generated using netcdf-4 with data subsetted using xaray, which I think should make the issues go away. This might make the `TimeParser` abstraction unnecessary. If so, we should remove it.

Remove the `TimeParser` stuff and go back to using a simple function, if possible.

charles-turner-1:
## Change Summary

The changes from #445 means we no longer need separate time parser abstractions per model, as this was purely the result of some dodgy test data. TLDR; I've reverted #293 as the previous approach was much simpler. 
## Related issue number

Closes #448 

## Checklist

- [ ] Tests pass on CI


marc-white (approved):
Code looks fine to me, although do we need to do an all-up build test before merging?

charles-turner-1:
Yeah, probably. E2E tests are currently broken following the atomic catalog build updates, so I'll fix those next week first. This PR is mostly spring cleaning anyway so there's no rush to merge.

charles-turner-1:
Blocked by #419

marc-white:
@charles-turner-1 do you want this into `main` now, or wait for #432 ?

charles-turner-1:
I think we should be fine to merge now tbh

PR #454, opened 2025-07-24
Add hash to _NCFileInfo

charles-turner-1:
## Change Summary

- Adds a hash of the datasets `._indexes` property to the `_NCFileInfo` class. 

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/src/access_nri_intake/source/builders.py
+++ b/src/access_nri_intake/source/builders.py
@@ -17,6 +17,7 @@ from .utils import (
     EmptyFileError,
     GenericTimeParser,
     GfdlTimeParser,
+    HashableIndexes,
     WoaTimeParser,
     _NCFileInfo,
     _VarInfo,
@@ -308,6 +309,7 @@ class BaseBuilder(Builder):
                 dvars.append_attrs(var, attrs)  # type: ignore
 
             start_date, end_date, frequency = cls.TIME_PARSER(ds, time_dim)()
+            index_hash = HashableIndexes(ds._indexes, [time_dim]).xxh
 
         if not dvars.variable_list:
             raise EmptyFileError("This file contains no variables")
@@ -319,6 +321,7 @@ class BaseBuilder(Builder):
             frequency=frequency,
             start_date=start_date,
             end_date=end_date,
+            index_hash=index_hash,
             **dvars.to_var_info_dict(),
         )
 
--- a/src/access_nri_intake/source/utils.py
+++ b/src/access_nri_intake/source/utils.py
@@ -3,9 +3,11 @@
 
 """Shared utilities for writing Intake-ESM builders and their parsers"""
 
+import json
 import os
 import re
 import warnings
+from collections.abc import Iterable
 from dataclasses import asdict, dataclass, field
 from datetime import timedelta
 from pathlib import Path
@@ -13,6 +15,9 @@ from pathlib import Path
 import cftime
 import polars as pl
 import xarray as xr
+import xxhash
+from frozendict import frozendict
+from pandas.api.types import is_object_dtype
 
 FREQUENCY_STATIC = "fx"
 
@@ -65,6 +70,7 @@ class _NCFileInfo:
     variable_standard_name: list[str]
     variable_cell_methods: list[str]
     variable_units: list[str]
+    index_hash: str
 
     def to_dict(self) -> dict[str, str | list[str]]:
         """
@@ -123,7 +129,7 @@ class _VarInfo:
     def to_var_info_dict(self) -> dict[str, list[str]]:
         """
         Return a dictionary representation of the _VarInfo object. Fields are
-        defined explicitly for use in the _AccessNCFileInfo constructor.
+        defined explicitly for use in the _NCFileInfo constructor.
         """
         return {
             "variable": self.variable_list,
@@ -132,6 +138,29 @@ class _VarInfo:
             "variable_cell_methods": self.cell_methods_list,
             "variable_units": self.units_list,
         }
+
+
+class HashableIndexes:
+    """
+    Consumes an xarray datasets `_indexes` attribute and creates a hashable representation.
+    Is this a bad idea? Probably
+    """
+
+    def __init__(self, _indexes: dict, drop_indices: Iterable[str] | None = None):
+        drop_indices = drop_indices or []
+        self.dict = frozendict(
+            {
+                key: tuple(val.index.to_list())
+                for key, val in _indexes.items()
+                if not is_object_dtype(val.coord_dtype) and key not in drop_indices
+            }
+        )
+
+        bytestream_dict = json.dumps(self.dict).encode("utf-8")
+        self.xxh = xxhash.xxh3_64(bytestream_dict).hexdigest()
+
+    def __repr__(self):
+        return str(self.xxh)
 
 
 class GenericTimeParser:
@@ -488,7 +517,6 @@ class AccessTimeParser(GenericTimeParser):
 
 
 class GfdlTimeParser(GenericTimeParser):
-
     def __init__(self, ds: xr.Dataset, time_dim: str):
         self.ds = ds
         self.time_dim = time_dim
--- a/tests/test_builders.py
+++ b/tests/test_builders.py
@@ -506,6 +506,7 @@ def test_generate_file_shape_info(builder, filename, time_dim, expected):
                 variable_standard_name=["", "", "", ""],
                 variable_cell_methods=["time: point", "time: point", "", ""],
                 variable_units=["degrees_N", "degrees_E", "degrees_E", "degrees_N"],
+                index_hash="71e155653bdbde78",
             ),
         ),
         (
@@ -555,6 +556,7 @@ def test_generate_file_shape_info(builder, filename, time_dim, expected):
                     "degrees_E",
                     "degrees_N",
                 ],
+                index_hash="2158d857cdf21b5f",
             ),
         ),
         (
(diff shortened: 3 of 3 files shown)


charles-turner-1:
@marc-white @dougiesquire I've requested reviews just so you get a ping, but this isn't quite ready to review yet - currently building a catalog against this branch.

charles-turner-1:
Interestingly, looks like this is increasing the jobfs usage - no idea why: 
```bash
...
/home/189/ct1163/access-nri-intake-catalog/bin/venv/lib/python3.11/site-packages/access_nri_intake/source/utils.py:212: UserWarning: Time coordinate does not include bounds information. Guessing start and end times.
  warnings.warn(
Job 145849693.gadi-pbs killed due to exceeding jobfs quota. Quota: 100.0MB, Used: 134.96MB, Host: gadi-cpu-clx-1916
/home/189/ct1163/access-nri-intake-catalog/bin/venv/lib/python3.11/site-packages/access_nri_intake/source/utils.py:212: UserWarning: Time coordinate does not include bounds information. Guessing start and end times.
  warnings.warn(
/home/189/ct1163/access-nri-intake-catalog/bin/venv/lib/python3.11/site-packages/joblib/externals/loky/backend/resource_tracker.py:299: UserWarning: resource_tracker: There appear to be 102 leaked semlock objects to clean up at shutdown
  warnings.warn(
/home/189/ct1163/access-nri-intake-catalog/bin/venv/lib/python3.11/site-packages/joblib/externals/loky/backend/resource_tracker.py:299: UserWarning: resource_tracker: There appear to be 12 leaked folder objects to clean up at shutdown
  warnings.warn(

======================================================================================
                  Resource Usage on 2025-07-24 20:44:34:
   Job Id:             145849693.gadi-pbs
   Project:            iq82
   Exit Status:        271 (Linux Signal 15 SIGTERM Termination)
   Service Units:      18.69
   NCPUs Requested:    48                     NCPUs Used: 48              
                                           CPU Time Used: 02:11:16        
   Memory Requested:   192.0GB               Memory Used: 135.26GB        
   Walltime requested: 06:00:00            Walltime Used: 00:11:41        
   JobFS requested:    100.0MB                JobFS used: 134.96MB        
======================================================================================
```

charles-turner-1:
As this is written, it'll add another field to our catalogs: `grid_id`, which is basically a hashed representation of the coordinate indices. 

I think we can build the `HashableIndexes` class into something that lets us represent the hierarchy of grids that @aidanheerdegen has been thinking about. To my mind it makes sense that tooling surrounding those grids would live in the catalog (at least until it gets sophisticated enough to break out to a separate package).

I'm envisaging adding something like  `__lt__`, `__gt__` methods which would let us determine whether one grid is a subset of another, although we'll probably need to have a bit of a think about whether we want that to represent an interpolation or reduced dimensions.

Anyway, the reason I'm soliciting opinions at this point is:
- This would add a new field to our ESM Datastores, which might confuse users (it might also be helpful?).
- I think it would start to firm up how we implement the [grids as a first class data product](https://forum.access-hive.org.au/t/grids-as-a-first-class-product/4722) idea, so I don't want to jump the gun and tie anyones hands too early.

marc-white:
> As this is written, it'll add another field to our catalogs: `grid_id`, which is basically a hashed representation of the coordinate indices.

I don't think there's anything inherently wrong with adding a new catalog field; I'm betting 90% of the users are likely not to notice.

However, how does this interact with the construction of datasets at catalog build time, and in particular, what's the interplay between `grid_id` and `file_id`? I don't see any updates to how things are being grouped together at build time...

charles-turner-1:
> > As this is written, it'll add another field to our catalogs: `grid_id`, which is basically a hashed representation of the coordinate indices.
> 
> I don't think there's anything inherently wrong with adding a new catalog field; I'm betting 90% of the users are likely not to notice.
>

I was hoping someone would say this, I suspect you're right...

> 
> However, how does this interact with the construction of datasets at catalog build time, and in particular, what's the interplay between `grid_id` and `file_id`? I don't see any updates to how things are being grouped together at build time...

Yeah, this is where it gets complicated, and in hindsight `grid_id` might have been a poor choice of name. So `file_id` essentially tells us about *the shape* of the grid and it's frequency - ie. `fx.x:300.y:300.z:50` for a static grid as an example.

The issue with that is that it doesn't encode any information about mergeability. As a dumb example, imagine you had two grids, both: `fx.x:180.y:180.z:50`. Under the current `file_id` system, we can have a couple of situations pop out:
1. Grid A. spans eg. [-180, 0), and grid B. spans [0, 180). These *are* concatenatable onto a grid spanning [-180, 180], and xarray will happily do this. However, the `file_id` of our merged grid would be `fx.x:360.y:180.z:50`, which gets a bit confusing.
2. Grid A. is defined by eg. `range(-180,180,2)` and grid B. by `range(-179, 181, 2)`. These two grids are still both of size (and `file_id` `fx.x:360.y:180.z:50`), and could in theory be interleaved onto one another, but xarray won't concatenate them - because it currently requires monotonic indices after concatenation (I don't think this will be true for long, BTW - the work to change this is mostly done).

In both these instances, the `grid_id` attribute would have different hashes for grid A and grid B, which could (would?) alert users to the potential for merging issues. The idea I was leaning towards was letting the `HashableIndexes` class handle this - whilst we'd just save the hashes out, the class lives in the `access_nri_intake` package, so users could import it & call some (as yet unfinished) utility functions to determine why their datasets might not be merging properly.

marc-white:
> Yeah, I think so. Basically, I think that `file_id` is a little too trigger happy when it comes to merging datasets (ie. the new `file_id`'s will wrongly & optimistically say datasets that xarray can't merge are the same on occasion) - but `grid_id` is a little too pessimistic about what can be merged. So in instances where loading on `file_id` fails, hopefully `grid_id` provides a way to disambiguate.

Do you propose to do this automatically, or is it the case that the user will need to intervene if a catalog build fails and manually set up a build based on `grid_id`? It might be quite a structural change to how the Builders work to do, "Try grouping by `file_id`, then if that doesn't work, try grouping by `file_id` and `grid_id`."

charles-turner-1:
Ahh, sorry, I could have been clearer. 

Because we don't attempt to load all our unique datasets into xarray datasets at build time, we don't really have any way of knowing if the `file_id` grouping will fail at build time. I've done a bit of investigating, and so far the only one that really seems problematic is the WOA one - because we have two grids (KDS, non-KDS), which have different (and unmergeable due to interleaving) vertical levels, but are the same size. There are a couple of others with 
 more `grid_id`'s than `file_id`'s I haven't properly checked though.

With that said, some instances with more `grid_id`'s than `file_id`'s are in fact mergeable. The `HI_nl_C_05_r1` contains a `time_0` coordinate that causes us to have almost 2000 `grid_id`'s but only 11 `file_id`'s - see picture
Screenshot 2025-07-29 at 11 03 08 am

Maybe we could extend the `BaseBuilder.parse_ncfile` method to allow us to specify multiple `time_dim`'s/, `drop_dims` and then stick that in the yaml files so that we can get the builder to check that all other dimensions that aren't dropped are concatenatable? Technically it's a bit dodgy as checking hashes is still too strict, but I think for the kind of output climate models write out it should be fine...

marc-white:
It's almost like you need to do a test Dataset creation for every experiment at catalog build time... which is in no way ideal or feasible.

How easy would it be to write instructions as to which dimensions should go in `time_dim`/`drop_dims`? E.g., if this dimension does "this", it needs to be included in this list of dimensions to drop.

charles-turner-1:
> It's almost like you need to do a test Dataset creation for every experiment at catalog build time... which is in no way ideal or feasible.
> 

Yup... 🫤 

> How easy would it be to write instructions as to which dimensions should go in `time_dim`/`drop_dims`? E.g., if this dimension does "this", it needs to be included in this list of dimensions to drop.

I wonder if we might be able to specify it in the config.yaml - eg, for the one I posted above, we could modify in [config/access-esm1-5.yaml](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/main/config/access-esm1-5.yaml),

with something like

```diff
  - metadata_yaml: /g/data/p73/archive/non-CMIP/ACCESS-ESM1-5/HI-nl-C-05-r1/metadata.yaml
    path:
      - /g/data/p73/archive/non-CMIP/ACCESS-ESM1-5/HI-nl-C-05-r1
    ensemble: false
+    extra_drop_dims: 
+      - time_0
```   

At least that way we hold it all in the build configuration. Presumably it's also something that we'd also expect to be consistent (ie. always occur for ESM1-5), so maybe we can just update the Builder to account for it.

charles-turner-1:
Ahh, I see what you're getting at. I think that we can potentially do something like this to help:

```python
@class_method
def parse_ncfile(cls,...):
    with xr.open_dataset(...) as ds:
        # Current code
        ...
        coord_hashdict = HashableIndices(ds, [time_dim])
        differing_dims = coord_hashdict ^ prev_hashdict # xor gives all dimensions where hashes differ
       if differing_dims:
           raise SomeSortOfConcatenationWarning(xyz)
```

Bit clunky and we'd need to somehow track all the hashes for a given dataset at once, but I think it should be doable.

marc-white:
That seems like a reasonable idea, although we'd need to be careful exactly where it was implemented - `parse_filename` is currently a `@class_method`, and unless I'm missing something, we'd need to do that on a per-instance basis.

charles-turner-1:
I think that actually might be kind of close to the desired behaviour - that way, the hash would exist on the class & it would get updated by every instance it touched, making sure it tracks the global state shared between all the files. 

I guess we would need to do something to differentiate between the hashes associated with each `file_id` though

PR #455, opened 2025-07-28
437 IPython magic to detect missing project

### LINKED ISSUE #437: Handling Missing Access to NCI Project Data More Gracefully
It would be great if, when users try to load data from an NCI project they do not have access to, they received a clear message explaining that the data is unavailable because they need to request project membership first.

At the moment, I think it is a bit confusing, it can throw a long error traceback that is hard to make sense of, especially for users who are not familiar with this kind of thing.

charles-turner-1:
## Change Summary

- [ ] Add IPython magic to detect missing project access
- [ ] Add tests
- [ ] Documentation?
## Related issue number

Closes #437

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/src/access_nri_intake/utils.py
+++ b/src/access_nri_intake/utils.py
@@ -1,15 +1,20 @@
 # Copyright 2023 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
 # SPDX-License-Identifier: Apache-2.0
 
-"""General utility functions for access-rni-intake"""
+"""General utility functions for access-nri-intake"""
 
+import ast
 import json
+import re
 from importlib import resources as rsr
 from pathlib import Path
+from typing import Any
 from warnings import warn
 
 import jsonschema
 import yaml
+from IPython.core.getipython import get_ipython
+from IPython.core.interactiveshell import ExecutionInfo
 
 from . import CATALOG_LOCATION, USER_CATALOG_LOCATION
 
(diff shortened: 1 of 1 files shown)


src/access_nri_intake/utils.py
```diff
@@ -153,3 +158,138 @@ def get_catalog_fp(basepath=None):
...
+        """Try to evaluate a node, or return the unparsed node if that fails.
+
+        Might be unnecessary.
+        """
+        try:
+            return ast.literal_eval(node)
+        except (ValueError, SyntaxError):
+            return ast.unparse(node)
+
+    def visit_Call(self, node: ast.Call) -> None:
```
marc-white, line 262: How difficult are the `ast.Call` objects to mock for testing purposes?

charles-turner-1: Ahh, good point. We could actually address that just by moving the imports into the accessor for the `intake.cat.access_nri` entrypoint.

charles-turner-1:
@rbeucher can you take a look at the demo notebook and tell me if this is roughly what you had in mind?

rbeucher:
Sure will do

marc-white:
@charles-turner-1 has something gone missing from the first cell of the demo notebook? It's blank, and my system can't locate the `%%check_load_call` magic.

marc-white:
> Sorry, yeah, it did. Should be fixed now.

Still looks blank to me, and looking at the underlying JSON, the `source` of that cell is still empty...

charles-turner-1:
Ahh - so pre commit doesn't recognise cell magics and deleted the import. Hopefully fixed now.

jemmajeffree:
I think that's the right spot

but. I'm concerned that nobody else will wade all the way through to the end and find it, especially as the relevant line is at the end of a very long error message.

Does this cell magic work on anything or just intake?

charles-turner-1:
> Does this cell magic work on anything or just intake?

Just intake - it's designed specifically to ignore anything else. 

> Actually, in general, I wonder if we could put a bunch of example cells right up the top
> 
> * load module
> * open catalogue
> * find some data
> * load this data (THIS CELL HAS THE CELL MAGIC LINE IN IT TOO)
> 
> Without any of the cell outputs, and with hyperlinks to the more detailed sections. That way people can easily click through to the bit they're strugling with, because the tutorial has a lot of detail in all the sections that makes it somewhat difficult to scroll through

Yeah, that is a good point. I'm not sure how hyperlinks to other cells work (if at all), but I'll see what I can figure out.

charles-turner-1:
@jemmajeffree @CodeGat Warm up AST for ya!

joshuatorrance (approved):
@charles-turner-1 has walked me through the code and it seems to do what we need it to do

PR #477, opened 2025-08-19
319 data request add aus2200 datasets

### LINKED ISSUE #319: [DATA REQUEST] Add AUS2200 datasets
# Description of the data product

AUS2200 is a high-resolution regional atmospheric model configuration that covers the entire Australian continent and surrounding oceans at 2.2km grid spacing, using the UM atmospheric model. This collection consists of multiple experiments of interest to the broad community. These include simulations of high-impact weather events such as major bushfire weather events and record-breaking flooding events, as well as multi-month simulations capturing important modes of tropical variability.

# Location of the data product on Gadi

/g/data/bs94/AUS2200/

# Checklist

Add a "x" between the brackets to all that apply
- [X] This data product is stable (unlikely to change substantially or move)
- [X] This data product is of use to the broader community
- [X] This data product is documented: <(https://geonetwork.nci.org.au/geonetwork/srv/eng/catalog.search#/metadata/f6014_0604_9188_1923)>
- [X] This data product is licensed under <CC-BY-4.0>
- [X] Those who want to access this data can be added to the project that houses it

joshuatorrance:
## Change Summary

This is still a work-in-progress.
- Added translator for AUS2200 dataset.

## Related issue number

This will eventually Close #319

## Checklist

- [x] Tests pass on CI
- [x] The new translator has been added to `__all__` in `src/access_nri_intake/catalog/translators.py`
-  [x] The new datastore has been added to the list of configs in `bin/build_all.sh` and `bin/test_end_to_end.sh`, and the storage flags in both scripts updated, if necessary.
- [x] You have generated a valid `metadata.yaml` for the datastore & placed it in the correct location: eg. `config/metadata_sources/aus2200/metadata.yaml` for the `aus2200` datastore.
    - `config/metadata_sources/aus2200/metadata.yaml`
- [x] The `metadata.yaml` has been copied to the correct location in `/g/data/xp65/admin/intake` -  for example, `/g/data/xp65/admin/intake/metadata/esgf-ref-qv56/metadata.yaml` for the `esgf-ref-qv56` datastore
    - `/g/data/xp65/admin/intake/metadata/aus2200/metadata.yaml` 
- [x] NCI datastore is in .prerelease dir
    - `/g/data/bs94/catalog/.prerelease/v2/esm/`
    - This may need to be moved at a date

## Notes

- In order to build a new catalog including this additional datastore, you will need to either release a new version of the `access-nri-intake` package, or run the `bin/build_all.sh` script to build a new catalog, using a virtual environment with the most recent (ie. including this PR) version of `access-nri-intake` installed. This can be done by activating the `venv` in `/g/data/xp65/admin/access-nri-intake-catalog/bin/build_all.sh` job.


src/access_nri_intake/catalog/translators.py
```diff
@@ -728,6 +728,8 @@ def _translate(string: str) -> tuple[str, ...]:
             "landonly": "land",
             "ocnBgChem": "ocnBgchem",
             "seaice": "seaIce",
+            "atmos": "atmos",
+            "land": "land",
```
charles-turner-1, line 732: I don't think these are necessary, the translator should do something like 

`realmdict.get(realm,realm)`, so cases where the key and value are the same shouldn't need to be defined.

src/access_nri_intake/catalog/translators.py
```diff
@@ -738,3 +740,50 @@ def _translate(string: str) -> tuple[str, ...]:
...
+        self.set_dispatch(
+            input_name="realm",
+            core_colname="realm",
+            func=super()._realm_translator,
+        )
+        self.set_dispatch(
+            input_name="frequency",
+            core_colname="frequency",
+            func=super()._frequency_translator,
+        )
```
charles-turner-1, line 782: I think these maps where we're using eg `super()._model_translator` and the names are the same don't need to be explicitly included - should be inherited from the base class.

Maybe double check? 1. I could be wrong and 2. It's actually kinda nice to have everything declaratively in the init

charles-turner-1 (commented):
Couple of minor comments, otherwise generally looks good!

src/access_nri_intake/catalog/translators.py
joshuatorrance: Right again!

I've removed the unnecessary code and popped in a comment as a compromise.

config/aus2200.yaml
```diff
@@ -0,0 +1,9 @@
+builder: null
```
joshuatorrance, line 1: Is builder being null okay for now? I presume this means that the datastore won't be automatically rebuilt?

charles-turner-1 (approved):
Looks good to me! 

Rest of the checklist stuff isn't in the repo, so I think we're ready to merge.

tests/test_translators.py
```diff
@@ -347,6 +348,25 @@ def test_CcamTranslator(test_data, groupby, n_entries):
...
+@pytest.mark.parametrize(
+    "groupby, n_entries",
+    [
+        (None, 100),
+        (["variable"], 23),
+        (["frequency"], 4),
+        (["model"], 1),
+        (["realm"], 2),
+    ],
+)
```
charles-turner-1, line 360: If we could, it would be great to minify this a touch further. I think if we:
1. Get unique variables
2. Subset df for first of each
3. Keep one of the thrown away rows

we should wind up with 24 lines, 23 variables and everything else unchanged?

Not super important, but it helps to keep the repo size & CI time down.

PR #478, opened 2025-08-25
489 - OM3 Builder can now extract realm from file attributes

joshuatorrance:
## Change Summary

- Realm has been added to _NCInfo and will be extracted from nc files if present in global attributes.
- OM3 builder will use realm from _NCInfo if filename doesn't match any of the defined patterns.
- This change should have no affect on existing builders/datasets unless there happens to be OM3 files being excluded that have realm defined.

## Related issue number

This change should close 489 if `realm` is added to the problem files' metadata (script to assist with this coming soon).

## Checklist

- [X] Unit tests for the changes exist
- [X] Tests pass on CI
- [X] Documentation reflects the changes where applicable


tests/test_builders.py
```diff
@@ -2965,3 +2965,24 @@ def test_builder_include_exclude_patterns(
...
         exclude_patts if exclude_patts is not None else default_exclude
     )
+
+
+def test_builder_om3_realm(test_data):
+    # om3-with-realm/ocean_month.nc has had global attr 'realm' set to 'ocean'
+    # om3-with-realm/ocean_month_no_realm.nc does not and should be invalid
+    #     om3 builder otherwise doesn't know the realm for that filename
+    # om3-with-realm/ocean_month_multiple_realms.nc has "ocean atmos"
+    data_path = str(test_data / "om3-with-realm")
```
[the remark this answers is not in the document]
joshuatorrance: Looks like it's expected to be a string.

If I leave it as a Path -
`TypeError: 'PosixPath' object is not iterable`
`ecgtools/builder.py:128: TypeError`

[ecgtools' builder](https://github.com/ncar-xdev/ecgtools/blob/8945f332e0b2dd5cc74e77ae7acbd2b818ec401e/ecgtools/builder.py#L111C5-L111C28) has `paths: typing.List[str]`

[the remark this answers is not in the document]
charles-turner-1: Cool, good to know!

tests/test_builders.py
```diff
@@ -2965,3 +2966,42 @@ def test_builder_include_exclude_patterns(
...
+    The .nc files have the 'realm' global attribute set and the filenames do not
+    match any of the normal patterns for extracting 'realm'.
+
+    Currently the multiple realms tests are expected to fail to parse since the
+    schema only allows a single realm.
+
+    FIXME: If the multiple realms file is not the first one to be parsed then the
+    building succeeds. This can be engineered by putting a 'z' in front of the
+    dirname.
+    """
```
charles-turner-1, line 2993: This seems kinda weird - do you think it might be a bug in the existing codebase?

joshuatorrance: Just for posterity's sake - Charles and I had a chat on this, looks like only the first valid file has its schema checked by [validate_parser](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/6b4b19762077dd5c33fd983fa781c4d26fcff39e/src/access_nri_intake/source/builders.py#L185).

For this PR we'll assume this is all working as intended. Tests will now only have files with correct schema. Might open up a new issue to discuss this later.

src/access_nri_intake/source/builders.py
```diff
@@ -511,7 +517,9 @@ def parser(cls, file) -> dict:
             elif "cice" in ncinfo_dict["filename"]:
                 realm = "seaIce"
             else:
-                raise ParserError(f"Cannot determine realm for file {file}")
+                # Default/missing value for realm is "" which is Falsy
+                if not (realm := str(ncinfo_dict["realm"])):
```
charles-turner-1, line 521: Didn't we reckon that `output_ncfile.realm` should get us round the type cast to string thing?

charles-turner-1 (commented):
Just the one comment and can we crunch down the size of those nc files a bit & I think we're ready to merge!

charles-turner-1 (approved):
Looks good!

Issue #168, opened 2024-05-02
ACCESS-CM3 Builder


dougiesquire:
ACCESS-CM3 is now at the point of producing output. It could be useful to have an ACCESS-CM3 Builder to allow us to build datastores for ACCESS-CM3 data.

ping @MartinDix, @kieranricardo

anton-seaice:
@sfiddes would also like a builder for UM regional nesting suite output - its not clear to be if this is the same or a different builder ? If someone is looking at this soon she can provide some example output :)

marc-white:
I think I'm on the hook for new builders @anton-seaice , so if you can provide some demo data I can start looking into it.

dougiesquire:
@paolap set up a builder (in a fork?) for Aus2200 output that could do what we need. Paola, are you planning on contributing that in to the main package?

sfiddes:
I have put a couple of files here: /scratch/public/slf563/UM_RNS. It is fairly niche sorry, but if there is something that works for AUS2022 that would be a good start... Also, I've just used iris to convert from pp files, so the naming might also be a bit off. I am looking for a better way to do the data processing to use CF conventions etc.. So perhaps not the best dataset to start with - AUS2022 probably better, but with recognition that not everyone uses AUS2022 :)

paolap:
> @paolap set up a builder (in a fork?) for Aus2200 output that could do what we need. Paola, are you planning on contributing that in to the main package?

Yes that was the idea, I was waiting to have time to review what I did and make it more flexible so that a user could control which extra fields would appear in the catalogue attributes.  I'm busy all this week and I would need to adapt it to whatever change Marc introduced,  I can have another look at it next week and than set a pull_request I did put  a link in another issue.

I also had a chat with Sonya previously about post-processing her data as Mopper has already the capacity of doing so, but currently starting from the raw model output rather than from the iris version. One issue with iris is that seems to make up standard_names, which is a potential source of information for mopper, so it would confuse the tool. Again I won't have anytime to look at this until next week.
Finally if there are examples of CM3 output I would love to get a sample so we can provide mapping for this version in our tool.

marc-white:
@paolap how did you go with your new builder?

paolap:
I didn't do anything more, maybe next week I'll have some time to review it and pull your changes. It was working for me for the example I wanted to run. The catalogue I produced is the one Navid used in his atmospheric cookbook example.

paolap:
[...]

PR #498, opened 2025-09-25
CM3 Builder

charles-turner-1:
## Change Summary

- Add CM3 Builder
- Add tests

## Related issue number

#168 

## Checklist

- [ ] Tests pass on CI


charles-turner-1:
Coverage appears to be getting confused here, I'm not sure exactly why, I think possibly the (now nonlocal) try-catch.

Sticking breakpoints or raising errors in the arms of the `if/else` chain in the Cm3 parser does trigger their execution, so these lines are being executed during testing.

Trying to work out why now, but probably unimportant...

PR #510, opened 2025-10-02
Cmip6Builder Ensemble bugfix

charles-turner-1:
## Change Summary

Fixes to address issues [on this forum post](https://forum.access-hive.org.au/t/building-intake-catalog-parser-returns-no-valid-assets-error/5320/11)

TLDR; updates the `Cmip6Builder` to separate datasets on `member` if the ensemble kwarg is passed as true. Reasoning being that ensemble runs of a model will all be on the same grid (and so mergeable), but should not be merged as part of the same dataset.

## Related issue number

None, but see forum post

## Please add any other relevant info below:

This is something we never really noticed as an issue before, but I think the changes to mergeability based `file_id` have created it. This will probably also need to be updated in the `AccessEsm15Builder`.

@dougiesquire can you confirm I'm barking up the right tree with the intent of the ensemble kwarg here?

### DIFF
--- a/src/access_nri_intake/source/builders.py
+++ b/src/access_nri_intake/source/builders.py
@@ -972,7 +972,7 @@ class Cmip6Builder(BaseBuilder):
     ]
     # PATTERNS is mostly unused here - we get the realm from the metadata. Only
     # using it to match on file names here & should be refactored to be removed. TODO
-    ensemble: bool = False
+    ensemble: bool = True
 
     def __init__(self, path, ensemble: bool, **kwargs):
         """
@@ -1012,6 +1012,7 @@ class Cmip6Builder(BaseBuilder):
                     "attribute_name": "member",
                 },
             ]
+            kwargs["groupby_attrs"] += ["member"]
 
         Cmip6Builder.ensemble = ensemble
 
@@ -1050,6 +1051,6 @@ class Cmip6Builder(BaseBuilder):
                         f"Cannot determine member for file {file} - "
                         "realization_index attribute missing"
                     )
-                ncinfo_dict["member"] = f"r{int(member_id):03d}"
+                ncinfo_dict["member"] = f"r{int(member_id)}"
 
         return ncinfo_dict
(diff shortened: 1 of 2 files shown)


dougiesquire:
Added to the todo list

src/access_nri_intake/source/utils.py
```diff
@@ -435,3 +436,14 @@ def _todate(t):
         frequency = frequency[1]
 
     return start_date, end_date, frequency
+
+
+@lru_cache(maxsize=10)
+def cache_xarray_open_dataset(*args, **kwargs):
```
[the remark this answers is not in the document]
charles-turner-1: Also, very nitpicky but I think `open_dataset_cached` might be a better name for this - I think it's slightly easier to read in our use case, eg.
```python
with cache_xarray_open_dataset(...
```
becoming
```python
with open_dataset_cached(...
```

Waddaya reckon?

src/access_nri_intake/source/builders.py
```diff
@@ -1047,4 +1058,19 @@ def parser(cls, file: str) -> dict:
             ]
         )
 
+        if cls.ensemble:
+            with cache_xarray_open_dataset(
```
joshuatorrance:
```suggestion
            with open_dataset_cached(
```

src/access_nri_intake/source/builders.py
```diff
@@ -365,9 +366,8 @@ def parse_ncfile(cls, file: str, time_dim: str = "time") -> _NCFileInfo:
 
         filename_frequency = cls.parse_filename_freq(file_path.stem)
 
-        with xr.open_dataset(
+        with cache_xarray_open_dataset(
```
joshuatorrance:
```suggestion
        with open_dataset_cached(
```

charles-turner-1:
I think I wrote this as a comment but it seems to have disappeared - I'm thinking it might be worth cherry-picking the caching stuff into a separate feature branch. 

NB. CI is failing because pixi doesn't seem to play nicely with commits straight from github. I'll try to work that one out ASAP.

PR #514, opened 2025-10-07
Handle corrupted/all nan time bounds

charles-turner-1:
## Change Summary

- The `_todate` function was only checking for whether a `bounds_var` existed, not whether it actually contained numbers. This updates that, setting `has_bounds` to False and falling back to using the `time` variable if that happens.
- Add tests.

Not sure quite sure where the issue arises from yet, but it was discovered by @ezhilsabareesh8 following on from https://github.com/ACCESS-Community-Hub/access-cm3-paper-1/issues/3#issuecomment-3358757408. 

@dougiesquire suggested 
> time_bnds looks okay in OM3 output. Possibly an inconsistency between the diagnostic output frequency and the run length?

## Related issue number

N/A, all in Zulip.

## Checklist

- [ ] Tests pass on CI


charles-turner-1:
From @blimlim: 

>In our somewhat hacky cleanup script, we're modifying the calendar for the MOM output https://github.com/ACCESS-NRI/dev_coupling/blob/67c611cd19bbafaffc438562d3ce346d74ce5f71/scripts/cleanup_output/cleanup_output.py#L53
> 
> I wonder if something could have gone wrong there?

PR #521, opened 2025-10-15
Cell methods fix - add `variable_cell_methods` to `groupby_attrs`

charles-turner-1:
## Change Summary

- Add `variable_cell_methods` to `groupby_attrs`
- Add test data, based on https://forum.access-hive.org.au/t/intake-bug/5398 by @julia-neme
- Add test based on above Julia's findings above
- (Unrelated) Add mypy to CI env - type checking stuff
___

This breaks some tests, because it changes number of datasets in datastores. This isn't necessarily a bad thing, but I think we need to be sure we are making the changes we want. See test notebook screenshow below:
Screenshot 2025-10-15 at 10 00 05 am

### DIFF
--- a/src/access_nri_intake/source/builders.py
+++ b/src/access_nri_intake/source/builders.py
@@ -209,7 +209,7 @@ class BaseBuilder(Builder, ABC):
             return cls.parser(file)
         except Exception:
             return set(
-                frozendict({INVALID_ASSET: file, TRACEBACK: traceback.format_exc()})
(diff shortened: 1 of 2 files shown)


dougiesquire:
One thing I was meaning to try and never got around to was to completely get rid of any columns with iterables, and instead just have multiple rows in the datastore for these assets. It would make the datastores much larger, and I'm not sure how well Intake-ESM would handle this off the bat, but the mods required to make this work could be quite simple.

Getting rid of columns with iterables would help with a lot of downstream headaches.

PR #557, opened 2026-01-08
Add "module use /g/data/xp65/public/modules" in quickstart notebook

aekiss:
## Change Summary

Add "module use /g/data/xp65/public/modules" in quickstart notebook (otherwise it won't work)

## Related issue number

N/A

## Checklist

N/A

### DIFF
--- a/docs/datastores/quickstart.ipynb
+++ b/docs/datastores/quickstart.ipynb
@@ -39,7 +39,7 @@
    "source": [
     "## Building an Intake-ESM datastore - the quick way\n",
     "\n",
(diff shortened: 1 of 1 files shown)


aekiss:
Could be good to also suggest using an interactive job, as a login node may have insufficient resources.

charles-turner-1 (commented):
Good call on the interactive job suggestion - I've added it as a suggestion (I'm hoping this lets me bypass the whole open a jupyter notebook, edit it there, commit work flow).

charles-turner-1 (approved):

Issue #539, opened 2025-11-16
Reserialize NCI Datastores into the build directory as parquet during build process


charles-turner-1:
# Is your feature request related to a problem? Please describe.

This will solve a couple of current issues:

1. Looks like it could be a while before those datastores will be serialised as parquet => performance 
2. The online catalog viewer tools[^1] assume that they can just scrape the build directory for the entire catalog. This means that they currently miss off anything that is translated and not built.

# Describe the feature you'd like

When we translate datastores, we should copy them across to our build directory, and reserialise them as parquet. I think that we open them anyway during the build process - we can just call `.serialize` whilst they're open & this should do the trick.

[^1]: https://reporting-dev.access-nri-store.cloud.edu.au/intake/catalog/ & https://charles-turner-1.github.io/catalog-viewer-spa/#/

PR #562, opened 2026-01-16
Take 2: Add configurable build via parquet (semi reverts #535, configurable, which reverted #518)

### LINKED ISSUE #540: Catalog selection on initialisation based on versioning

We recently reserialised the catalog as parquet for read time & storage size optimisations. However, older versions of intake-esm don't have the ability to read parquet. This causes utf8 decode errors, as they try to open a parquet file as a CSV.

Since we need to support old environments until they are deprecated for other reasons (ie. zarr 3/ PARCELS compatibility), we need some way of supporting this.

Newer versions of the catalog should detect that they are able to read parquet & open a parquet version of the catalog. Older versions should be unaware of the parquet formatted catalog, and just read a CSV mirror of it. Probably we'll want to keep the csv version gzipped - both for storage size, but also because it being slow might(?) encourage people to update their environment.

Just keep everything as CSV until we full deprecate old environments.

charles-turner-1:
## Change Summary

- `CatalogManager` class can be configured with the `use_parquet` flag to serialise the catalog as parquet.

## Related issue number

- [x] #518 
- [x] #535 
- [x] #540
- [ ] #539 || I'd like to hit this in this PR, but I think it's unfeasibly large to include the changes in here.

## Checklist

- [x] Tests pass on CI
- [ ] Documentation reflects the changes where applicable N/A

___

Closes #540 

N.B This will also introduce a pile of bugs in the `data.utils` module surrounding versioning information available to users. TBH I don't confident addressing those in this PR: the version logic has gotten so out of hand with this (notes in the code comments) that we need to do some major refactoring before any of that can be addressed IMO.


joshuatorrance (approved):
This is a big PR, I haven't spotted anything problematic though

PR #564, opened 2026-01-19
Reserialise NCI datastores to parquet in catalog build location

### LINKED ISSUE #539: Reserialize NCI Datastores into the build directory as parquet during build process

charles-turner-1:
Closes #539 

## Checklist

- [x] Tests pass on CI
- [ ] Documentation reflects the changes where applicable N/A


joshuatorrance (approved):
LGTM

PR #587, opened 2026-03-20
586 Fix deploy workflow

charles-turner-1:
## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/.github/workflows/gadi_deploy.yml
+++ b/.github/workflows/gadi_deploy.yml
@@ -10,11 +10,8 @@ on:
       concretise:
         description: 'Whether to concretise the catalog (move from .vYYYY-MM-DD to vYYYY-MM-DD and update default catalog.)'     
         required: true
-        default: 'true'
-        type: choice
-        options:
-          - true
-          - false
+        default: true
+        type: boolean
       format:
         description: 'Whether to serialise catalogs as parquet or csv'
         required: true
--- a/.github/workflows/gadi_deploy.yml
+++ b/.github/workflows/gadi_deploy.yml
@@ -4,12 +4,38 @@ on:
     inputs:
       release_version:
         description: 'Release version'     
+        type: string
+        default: ''
+        required: false # Not required
+      concretise:
+        description: 'Whether to concretise the catalog (move from .vYYYY-MM-DD to vYYYY-MM-DD and update default catalog.)'     
         required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/gadi_deploy.yml
```diff
@@ -4,12 +4,38 @@ on:
     inputs:
       release_version:
         description: 'Release version'     
+        type: string
+        default: ''
```
CodeGat: This is the default default, so this can be removed.

.github/workflows/gadi_deploy.yml
```diff
@@ -4,12 +4,38 @@ on:
...
+        required: false # Not required
+      concretise:
+        description: 'Whether to concretise the catalog (move from .vYYYY-MM-DD to vYYYY-MM-DD and update default catalog.)'     
         required: true
+        default: 'true'
+        type: choice
+        options:
+          - true
+          - false
```
CodeGat: Rather than a list of `true` or `false`, you can use `boolean` instead:


```suggestion
      concretise:
        description: 'Whether to concretise the catalog (move from .vYYYY-MM-DD to vYYYY-MM-DD and update default catalog.)'     
        required: true
        default: true
        type: boolean
```

.github/workflows/gadi_deploy.yml
```diff
@@ -18,8 +44,21 @@ jobs:
           key: ${{secrets.DEPLOY_KEY}}
           script: |
             cd ${{secrets.GADI_REPO_PATH}}
-            export RELEASE=${{ github.event.inputs.release_version }}
+            git pull && git checkout ${{ github.ref_name }}
```
CodeGat: Just noting that this will most likely always be `main` unless you change the "Run workflow from..." bit in the `Run Workflow` box.

CodeGat (commented):
Just a couple commentss

charles-turner-1:
Thanks! 🙏

.github/workflows/gadi_deploy.yml
CodeGat: I also don't think this input is used (yet)

CodeGat (commented):
The below can be shortened, but outside of that it looks fine :)

CodeGat (approved):

PR #599, opened 2026-04-07
AccessOm2Builder: change default to file realm metadata

charles-turner-1:
- Update ordering in AccessOm2Builder.parser() so that we default to file metadata realm & don't break if we can't regex it
- Also simplify temporal label default in utils (strings are immutable)

I deliberately haven't added any tests yet because I want to see what codecov says - this was previously a logic bug I'm a little surprised our tests didn't cover, although it's fairly subtle.

__I haven't checked the other parsers for this kind of thing. We should probably do that.__

## Checklist

- [x] Tests pass on CI


joshuatorrance (approved):
LGTM

PR #600, opened 2026-04-13
Widen builder include patterns

charles-turner-1:
## Change Summary

Scope TBC. Might be targeted, might be a bit more sweeping. 

Following https://forum.access-hive.org.au/t/datastore-making-unidentified-realm/6216/12

The basic idea: we should be parsing most of the details out of the netCDF's anyway now, so we might as well make the include patterns as broad as possible.

## Checklist

- [NA] Unit tests for the changes exist
- [x] Tests pass on CI

### DIFF
--- a/docs/management/source.rst
+++ b/docs/management/source.rst
@@ -55,6 +55,11 @@ should be required is to:
   <https://ecgtools.readthedocs.io/en/latest/how-to/use-a-custom-parser.html>`_ may be useful. Again, it will 
   also be helpful to look at existing Builder class implementations.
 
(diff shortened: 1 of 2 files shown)


joshuatorrance (approved):
LGTM

Were the PATTERNS even used anywhere?

charles-turner-1:
So I thought they were being used eg. [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/8703e1853e1feb5118845b98bb7caf20f333e49c/src/access_nri_intake/source/builders.py#L447), but nah, it looks like they weren't being used anywhere at all...

PR #602, opened 2026-04-14
No longer require long name to be present & fix related test parametrisations

charles-turner-1:
## Change Summary

Following https://forum.access-hive.org.au/t/datastore-making-unidentified-realm/6216/15, updated the netCDF parsing to no longer require parsers have a long name.

This requirement seems to be totally vestigial: early versions of this package had something like

```
if long_name in attrs:
   nc_info["long_name"] = attrs["long_name"]
```

which got turned into a(n unnecessary) guard clause during a couple of refactors:

```
if long_name not in attrs:
   return None
```

I'm pretty sure it's safe to remove completely.

## Related issue number

None, but see forum issue

## Checklist

- [x] Tests pass on CI


dougiesquire (approved):
LGTM @charles-turner-1

PR #616, opened 2026-05-11
Chore: update lockfile to v7

charles-turner-1:
as per title.
```
$ pixi shell
 WARN the lock file is up-to-date but uses an older format (v6), run `pixi lock` to upgrade to v7 for improved reproducibility
$ pixi lock
 WARN the lock file is up-to-date but uses an older format (v6), re-solving all environments using locked content to upgrade to v7
✔ Updated lock-file
```


joshuatorrance (approved):
LGTM

PR #642, opened 2026-07-08
`build_all.sh`: `--use-parquet` on by default

joshuatorrance:
## Change Summary

- `build_all.sh` now has `--use-parquet` in the command string.
- Tweaked handling of command string to avoid duplication

## Related issue number
N/A


charles-turner-1 (approved):
LGTM!

PR #643, opened 2026-07-13
Warn if we run use-datastore on a login node

charles-turner-1:
## Change Summary

Warn if we run the build datastore CLI from a login node. We might want to capture and reraise with this warning if the scheduler does kill the job too?

## Related issue number
See https://forum.access-hive.org.au/t/build-esm-datastore-failing-in-conda-analysis3-26-03/6245/21

 > Just to note, build-esm-datastore still fails (slowly, with prolific and scary-looking error messages) on a login node in conda/analysis3-26.04 through to conda/analysis3-26.07. But ARE works.
>
> Maybe it would be better for build-esm-datastore to simply abort with a message to use ARE if run on a login node?

## Checklist

- [x] Tests pass on CI

### DIFF
--- a/src/access_nri_intake/cli.py
+++ b/src/access_nri_intake/cli.py
@@ -25,7 +25,7 @@ from .data import CATALOG_NAME_FORMAT
 from .experiment import use_datastore
 from .experiment.colours import f_info, f_path, f_reset
 from .experiment.main import scaffold_catalog_entry as _scaffold_catalog_entry
-from .experiment.utils import parse_kwarg, validate_args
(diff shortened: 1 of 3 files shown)


charles-turner-1:
I'm just looking into whether we can use a try/catch to swallow these errors and re-emit with a 'killed by scheduler' error too. Not totally sure it'll be possible (SIGINT/SIGKILL), but we'll see

charles-turner-1:
@charles-turner-bot verify you're able to respond here please

charles-turner-1:
@charles-turner-bot again

PR #649, opened 2026-08-07
Fix parsing func kwargs, enable ensembles for AccessEsm16Builder

charles-turner-1:
## Change Summary

Supersedes #648 

- Threads `parsing_func_kwargs` though any of the builders that needs them. Because of the joblib.parallel stuff in ecgtools, classes need to be stateless (pickling/unpickling correctly), so parsing funcs get passed in though this class, not via hacks to modify class attributes in `__init__`. (We were doing this previously because I hadn't understood the layer below).
- Fixes a couple of bugs where the ESM1.5, ESM1.6 and CM2 builders weren't handling parsing funcs correctly due to the above.
- Fixes a couple of tests and makes the test data a bit more precise.

## Related issue number

- [ ] Needs checking

## Checklist

- [x] Tests pass on CI

___ 

## From my bots description:
- thread `parsing_func_kwargs` through `BaseBuilder` to the ecgtools parser wrapper
- apply the explicit parser-kwarg pattern to the ACCESS-ESM ensemble builders as well as `Cmip6Builder`
- remove parser-time dependence on mutable class state while preserving existing test behaviour


src/access_nri_intake/source/builders.py
```diff
@@ -729,7 +733,9 @@ class AccessCm2Builder(AccessEsm15Builder):
 class AccessEsm16Builder(AccessEsm15Builder):
     """Intake-ESM datastore builder for ACCESS-ESM1.6 datasets"""
 
-    PATH_REGEX = r".*/output\d+/([^/]*)(?:/[^/]*)?/.*\.nc"
+    PATH_REGEX = (
+        r".*/[^/]*-(r\d+i\d+p\d+f\d+)-[0-9a-f]+/output\d+/([^/]*)(?:/[^/]*)?/.*\.nc"
```
joshuatorrance: Perhaps something like this:
`.*(?:\/[^/]*-(?<realization>r\d+i\d+p\d+f\d+)-[0-9a-f]+)?\/output\d+\/(?<realm>[^/]*)(?:\/[^/]*)?\/.*\.nc`

Also some comments/examples might help people down the road in parsing the regex

tests/test_builders.py
```diff
@@ -317,15 +342,40 @@ def test_builder_build(
...
+            "r10i1p1f1",
+            "xt_ocean:2.yt_ocean:2",
+        ),
+        (
+            "access-esm1-6/historical-historical-1.1-r10i1p1f1-aaca3142/historical-historical-1.1-r10i1p1f1-aaca3142/output001/ocean/ocean-2d-roughness_amp.nc",
+            "AccessEsm16Builder",
+            "ocean",
+            None,
+            "xt_ocean:2.yt_ocean:2",
+        ),
```
charles-turner-1, line 358: @joshuatorrance these are where we'll put the regexes in - we'll need to grab and minify the files too

joshuatorrance:
Here's a couple more ESM1.6 datasets that could be used to create test cases:
- Postprocessed ESM1.6: `/scratch/tm70/jt4085/esm1p6_postprocessing/1pctCO2-01/output000/atmosphere/access-esm1p6.um7p3.3d.fld_s33i002.1mon.mean.0201.nc`
- Not postprocessed ES1.6: `/scratch/tm70/jt4085/esm16_examples/1pcC02-01/output000/atmosphere/netCDF/aiihca.pa-020101_mon.nc`

charles-turner-1:
Think this should be a lot slicker now. If we can't spot any more holes, I'll tack this approach onto all the other builders too I think...

joshuatorrance (approved):
LGTM!

PR #653, opened 2026-08-11
Bring all other builder parsers up to spec with ESM1.6 Builder, where relevant

### LINKED ISSUE #650: Clean up builder regex parsing
We should follow the style introduced by #649 in all builders, including probably putting an explanation of the regex in the docstrings for each builder.

charles-turner-1:
Closed #650, follows template of #649

### DIFF
--- a/src/access_nri_intake/source/builders.py
+++ b/src/access_nri_intake/source/builders.py
@@ -721,7 +721,7 @@ class AccessEsm15Builder(BaseBuilder):
         )
 
     @classmethod
-    def parser(cls, file, ensemble: bool = True) -> dict:
+    def parser(cls, file, ensemble: bool = False) -> dict:
         nc_info = cls.parse_ncfile(file)
         ncinfo_dict = nc_info.to_dict()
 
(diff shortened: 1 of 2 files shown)


src/access_nri_intake/source/builders.py
```diff
@@ -694,24 +706,30 @@ def __init__(self, path, ensemble: bool, **kwargs):
...
+        """
+        match = re.match(cls.PATH_REGEX, file)
         if not match:
-            raise ParserError(
-                f"Unable to parse filepath {file} in {cls.__class__.__name__}"
-            )
-
-        exp_id = match.groups()[0]
-        realm = match.groups()[1]
+            raise ParserError(f"Unable to parse filepath {file} in {cls.__name__}")
```
joshuatorrance, line 715: If I'm reading CodeCov right this line is not covered by the tests, but the `raise ParserError` before was?

Can we add a quick test case that fails to parse?

joshuatorrance (approved):
LGTM

PR #658, opened 2026-08-17
Add first pass at AM3 builder

### LINKED ISSUE #657: Add AM3 builder
The AM3 model does not yet have a data builder. Should be able to use the existing CM3 builder as a template for the AM3 builder, since they use the same atmosphere model (although directory structures may be different).

Whyborn:
## Change Summary

This adds a builder for ACCESS-AM3, which is currently missing.

## Related issue number

Fixes #657 

## Checklist

- [x] Tests pass on CI

- [ ] The new translator has been added to `__all__` in `src/access_nri_intake/catalog/translators.py`
-  [ ] The new datastore has been added to the list of configs in `bin/build_all.sh` and `bin/test_end_to_end.sh`, and the storage flags in both scripts updated, if necessary.
- [ ] You have generated a valid `metadata.yaml` for the datastore & placed it in the correct location: eg. `config/metadata_sources/esgf-ref-qv56/metadata.yaml` for the `esgf-ref-qv56` datastore.
    - `<<REPLACE THIS TEXT WITH YOUR METADATA FILE IN THIS PACKAGE>>`
- [ ] The `metadata.yaml` has been copied to the correct location in `/g/data/xp65/admin/intake` -  for example, `/g/data/xp65/admin/intake/metadata/esgf-ref-qv56/metadata.yaml` for the `esgf-ref-qv56` datastore
    - `<<REPLACE THIS TEXT WITH THE PATH TO METADATA FILE IN ITS FINAL LOCATION ON DISK>>` 

## Notes

- In order to build a new catalog including this additional datastore, you will need to either release a new version of the `access-nri-intake` package, or run the `bin/build_all.sh` script to build a new catalog, using a virtual environment with the most recent (ie. including this PR) version of `access-nri-intake` installed. This can be done by activating the `venv` in `/g/data/xp65/admin/access-nri-intake-catalog/bin/build_all.sh` job.


Whyborn:
@paocorrales Here's the PR- I'm setting it up to test it out on Gadi.

Whyborn:
@paocorrales I think this is now testable if you want to give it a go- unfortunately we don't have a prerelease workflow for this, but you should be able to install this in a virtual environment and then run:
```
build-esm-datastore --builder AccessAm3Builder --expt-dir <path-to-run>
```
where `<path-to-run>` is the path to the top level of the run e.g. `/scratch/gb02/${USER}/cylc-run/access-am3-dev-n512e` (with the addition of the run name if it's a Cylc8 run).

Whyborn:
This appears to be functional. I'm having a bit of trouble following the test suite- @charles-turner-1 can you give a bit of guidance as to what I should be including in the `tests/test_builders.py`?

charles-turner-1:
> This appears to be functional. I'm having a bit of trouble following the test suite- @charles-turner-1 can you give a bit of guidance as to what I should be including in the `tests/test_builders.py`?

Do you mind if I push a couple of stubs into the test file straight to this branch? Probably gonna be the easiest way

Whyborn:
Yea go for it @charles-turner-1

Whyborn:
Ok, I've filled out the test template and added the test data. The test data used for the size checks is a bit on the larger size- I don't know why it's so much larger than the other examples in the folder, but it's still only a few MB.

The AM3 outputs don't have the frequencies encoded in the name yet- this will need to be updated when we coordinate the output stream names.

tests/test_builders.py
```diff
@@ -61,6 +61,7 @@
         (["woa"], "WoaBuilder", {}, 8, 8, 3),
         (["cmip6"], "Cmip6Builder", {"ensemble": False}, 74, 72, 14),
         (["cmip6"], "Cmip6Builder", {"ensemble": True}, 74, 72, 31),
+        (["access-am3"], "AccessAm3Builder", {}, 2, 2, 2),
```
charles-turner-1: This implies 2 valid assets, 0 invalid & 2 datasets I think. Can we add an invalid asset and another valid asset that should merge into one of the existing datasets to stress test the builder?

charles-turner-1 (commented):
One minor comment but looking good!

tests/test_builders.py
Whyborn: What classifies an invalid dataset? I think I should be able to add a dataset that merges.

Whyborn: It seems like most of our model builders don't include any invalid datasets in the test- all the NetCDF output should be includable I think.

Whyborn:
I've added some merging datasets, but no invalid assets- the model output doesn't come with any that should be considered invalid.

tests/test_builders.py
charles-turner-1: I think if you just create an empty nc file, ie. `touch junk.nc` it should come up as invalid

charles-turner-1 (approved):
LGTM.

Do we have outputs we want to add to the main catalog yet or are we still waiting on those?

Whyborn:
Yet to decide precisely what we want to put in the catalog, and where we should put it.

PR #659, opened 2026-08-17
Add xxhash as an explicit dependency

charles-turner-1:
## Change Summary

Was implicitly a dependency of yamanifest & had gone missing.

## Related issue number

https://github.com/ACCESS-NRI/om3-scripts/pull/155#issuecomment-5312122688)


joshuatorrance (approved):
LGTM

PR #663, opened 2026-08-21
Fix hive-partitioned datastore reads and update cordex schema

charles-turner-1:
## Change Summary

Bundles a few related changes to the cloud mirror (`src/access_nri_intake/cloud.py`) plus the cordex schema fix that started this branch.

**Fix hive-partitioned datastore reads.** `partition_parquet_files` writes any datastore listed in `PARTITION_TABLE` out as a hive-partitioned *directory* rather than a single `.parquet` file. Polars only reads a hive dataset when the path ends in a trailing slash, so `update_esm_datastores` now appends `/` to the JSON `catalog_file` for those datastores (keyed on `PARTITION_TABLE` membership); unpartitioned datastores are left untouched. This resolves the `TODO` in `partition_parquet_files`, which incorrectly pointed at the metacatalog — the parquet path polars opens lives in each datastore's JSON `catalog_file`, not the metacatalog `yaml` column (which only references the `.json` descriptor).

**Update cordex schema.** NCI changed the `cordex_ig45` schema from `experiment_id` to `driving_experiment_id`; `PARTITION_TABLE` is updated to match.

**Extract the bucket name.** The Nectar bucket name is now defined once as `BUCKET_NAME` / `BUCKET_BASE_URL` instead of being repeated as string literals.

**Add a `marimo` pixi environment** for interactive notebook work alongside the dev toolchain.

## Related issue number

N/A

## Checklist

- [x] Tests pass on CI


joshuatorrance (approved):
LGTM

PR #669, opened 2026-08-31
Update to experiment metadata v1.0.4

charles-turner-1:
@joshuatorrance 

All tests pass locally


charles-turner-1:
Tests should pass now. 

I suspect we'll have errant v1.0.3 metadata's on disk on Gadi. I'll a `no-concretise` build to track down any bugs.

I also haven't touched `metadata_schema_file.json`. One for the future - let's just fix this first to keep any potential build errors tractable.

src/access_nri_intake/data/metadata_schema_experiment.json
joshuatorrance: The current build will still fail for some of the new `jq44` `metadata.yaml` files that have an extra field, `jobname`.

For now I suggest we relax the schema:

```suggestion
    "additionalProperties": true
```

joshuatorrance (approved):
One suggestion (`additionalProperties`), otherwise LGTM

PR #671, opened 2026-08-31
Smarter building

### LINKED ISSUE #578: Efficient Catalogue Building (like make)

In theory, when re-building the catalogue we only need to rebuild datastores that are new or have changed. If we can come up with a way to figure out which datastores have not changed then we can copy them from the last successful build.

This could significantly speed up the build which would save SU and enable regular automated builds (e.g. daily, weekly).

There's a few pieces to this problem:
- How do we know if a dataset has changed since the last build?
- How do we know if a dataset is new?
- We need a way to control which datastores are created during a build
- We need a way to copy datastores from the previous build

joshuatorrance:
## Change Summary

Makes a few tweaks to how the catalogue is built:
- When building a datastore:
  - Check if the list of files has changed
    - Use the datastore to get a list of assets
      - INVALID assets are now cached to the datastore's parquet file
    - Use the Builder to get the list of all assets
      - We don't know which are valid yet
    - Compare the two lists
  - Check if the `mtime` on any of those files is newer than the `mtime` on the datastore

If the datastore doesn't need rebuilt then use the `load` function to load the datastore from the previous version.

## Related issue number

Closes #578 

## Checklist

- [x] Tests pass on CI


src/access_nri_intake/catalog/manager.py
```diff
@@ -149,6 +162,38 @@ def build_esm(  # noqa: PLR0913, PLR0917 # Allow this func to have many argument
...
+        if datastore_path.suffix == "parquet":
+            datastore = pl.read_parquet(datastore_path).to_pandas()
+        elif datastore_path.suffix == "csv":
+            datastore = pl.read_csv(datastore_path)
+        else:
+            # FIXME: need a more appropriate error here
+            raise FileExistsError(
+                f"Unexpected filetype for datastore: {datastore_path.suffix}"
+            )
+
```
charles-turner-1, line 196:
```suggestion
# Check if the number of files in the datastore is the same as the number of *valid assets* the builder reports. The valid assets bit is where the difficulty comes in.
if len(datastore) != num_valid_assets:
     return NEEDS_REBUILD
     
if tuple(datastore.path_column) != valid_assets:
    return NEEDS_REBUILD
```

src/access_nri_intake/catalog/manager.py
```diff
@@ -149,6 +162,38 @@ def build_esm(  # noqa: PLR0913, PLR0917 # Allow this func to have many argument
...
+        Parameters
+        ----------
+        datastore_path: Path
+            The path to the datastore to check
+
+        Returns
+        -------
+            True is the datastore needs to be rebuilt, otherwise False
+        """
```
charles-turner-1, line 185:
```suggestion
        """
        NEEDS_REBUILD = True
        USE_BUILT = False
```
We probably won't wanna do this IRL but for the sake of clarity here

src/access_nri_intake/catalog/manager.py
charles-turner-1: The tricky bit with this is that I haven't found a faster way of getting the number of, let alone the actual correct valid assets, than just building the datastore.

charles-turner-1: Breaking it down case by case:
1. Missing files in existing datastore:
   - len(datastore) < len(files_in_dirs)
   - But this isn't conclusive because there might be invalid assets
2. Extra files in datastore:
  - len(datastore) > len(files_in_dir)
  - This probably does tell us we need a rebuild, to be fair. 
3. A file that was valid is now invalid, for whatever reason. This shouldn't really be possible, but seems to happen occasionally. We can't determine this by the length, only by the valid_assets.

charles-turner-1:
Sorry to be so negative with those comments: I think the real issue (as I see it), is that we can't reliably work out whether a datastore *is currently valid*. It's much easier to work out whether it's *currently invalid*, which is a far less strict test. Unfortunately, we need to know that it's currently valid to skip building it.

One thing we might be able to try is an approach more like this:
https://github.com/ACCESS-NRI/intake-virtual-icechunk/pull/91

It's not really gonna be a like for like transfer in any meaningful sense, but the builders are opening everything with xarray just pull out variable names, etc. Creating these xarray objects has a lot of overhead - in the intake-virtual-icechunk PR I measured it at around 80% of the total runtime. We could probably adapt that from zarr to netcdf to get a pretty substantial performance boost.

src/access_nri_intake/catalog/manager.py
joshuatorrance: Great points!

1. Files that are no longer present would raise an exception when we try to get their mtime, that should be easy enough to catch and mark for rebuilding.
2. As you've already said this one is tricky... perhaps we can extract the file-finding portion of the build and run that early to do a count?
3. If a file hasn't changed since the last valid build it's probably pretty safe to assume it's still valid. We'll need to think about if and how we might want to track the builders changing.

joshuatorrance:
> ...we can't reliably work out whether a datastore _is currently valid_.

Do we need to? If a datastore was previously built and nothing has changed with the data or the builder then rebuilding achieves nothing.

charles-turner-1:
Maybe not, if we expose some sort of `--nuke-it` flag, to be fair. My major concern is more just propagating along invalid datastores.

I'm thinking of cases where, for example:
```
output/
  1900_01/
    file_1_jan.nc
    file_2_jan.nc
  1900_02/
    file_1_jan.nc
    file_2_jan.nc
```
becomes for example
```
output/
  1900/
    file_1_jan.nc
    file_2_jan.nc
    file_1_feb.nc
    file_2_feb.nc
```

The number of files is gonna be the same. I'm not 100% on how mtimes work, but I consulted professor GPT and it said that mtimes shouldn't be altered by an `mv` unless the files are moved across filesystems.

I'm struggling to think of a more pathological scenario off the top of my head right now to be fair.

I guess I'm just erring towards a setup where failed checks => rebuild rather than do nothing/ parse-don't-validate setup. Maybe I'm being overly cautious... thoughts?

joshuatorrance:
> Maybe not, if we expose some sort of `--nuke-it` flag, to be fair. My major concern is more just propagating along invalid datastores.
> 
> I'm thinking of cases where, for example:
> ...

Good example, rather than just checking the count of files my current draft checks the list of paths which should pick up changes like that. In brief, before checking mtimes, we check the list of paths in the datastore matches the list of paths the build is going to parse. As before nothing properly tested yet though.

src/access_nri_intake/catalog/manager.py
```diff
@@ -149,6 +165,45 @@ def build_esm(  # noqa: PLR0913, PLR0917 # Allow this func to have many argument
...
+        elif datastore_path.suffix == "csv":
+            datastore = pl.read_csv(datastore_path)
+        else:
+            # FIXME: need a more appropriate error here
+            raise FileExistsError(
+                f"Unexpected filetype for datastore: {datastore_path.suffix}"
+            )
+
+        # Check the assets the builder will build matches the those in the datastore
+        if sorted(datastore["path"].to_list()) != sorted(builder.get_assets().assets):
```
joshuatorrance, line 198: Do the .assets include the INVALID_ASSETS?

joshuatorrance: No they don't. :(

Need to actually go through the parsing process to identify invalid assets

joshuatorrance:
A test of a complete catalogue build with the current state of this PR took around 2h 30min - that's while rebuilding anything with invalid assets.

charles-turner-1:
Nice - the last pre-caching run took just under four hours, so that's a decent improvement!

Do you know what fraction of datastores we managed to skip rebuilding?

joshuatorrance:
> Nice - the last pre-caching run took just under four hours, so that's a decent improvement!
> 
> Do you know what fraction of datastores we managed to skip rebuilding?

109 datastore were reused, 9 were rebuilt

charles-turner-1:
Interesting - looks like we might be getting close to a theoretical floor for opening this many files then?

joshuatorrance:
I've added a test that fails with invalid assets but should succeed if we add a feature to handle those cases

PR #680, opened 2026-09-04
Allow multiple schema versions per build

charles-turner-1:
## Checklist

- [x] Tests pass on CI


src/access_nri_intake/cli.py
```diff
@@ -928,6 +937,8 @@ def metadata_validate(argv: Sequence[str] | None = None):
         if Path(f).is_file():
             print(f"Validating {f}... ")
             try:
+                schema_version = load_metadata_yaml(f, None).get("schema_version", None)
+                EXP_JSONSCHEMA = get_versioned_schema(schema_version)
```
joshuatorrance: Uppercase variable names are usually for constants variables right? Should we lowercase this name now?

src/access_nri_intake/utils.py
```diff
@@ -151,3 +155,21 @@ def get_catalog_fp(basepath=None):
...
+    """
+    Get the versioned schema for the catalog.yaml file. If the version is not found, return the newest schema.
+    """
+
+    if schema_version == "1-0-4":
+        metadata_file = "data/metadata_schema_experiment_1-0-4.json"
+    elif schema_version == "1-0-3":
+        metadata_file = "data/metadata_schema_experiment_1-0-3.json"
+    else:
+        metadata_file = "data/metadata_schema_experiment_latest.json"
```
joshuatorrance: We've already hardcoded the versions here, should we just hardcode which one is `latest` rather than duplicating a file?


```suggestion
    if schema_version == "1-0-3":
        metadata_file = "data/metadata_schema_experiment_1-0-3.json"
    else:
        # Default to the most recent version
        metadata_file = "data/metadata_schema_experiment_1-0-4.json"
```

charles-turner-1: I actually tried this & it was surprisingly messy. Somewhere else in the codebase we import the schema - doing it this way means we don't need to percolate the changes out there if we do update things.

joshuatorrance: Fair enough!

Do you think it's worth removing the hardcoded versions then? I.e. use glob to look for `data/metadata_schema_experiment_*.json` file, handle `latest` and order the versions automatically. Ordering versions is not the same as .sort unfortunately so I suspect it's not worth messing about at the moment.

Just thought of another slightly simpler variant on that if:
```suggestion
    if schema_version:
        metadata_file = f"data/metadata_schema_experiment_{schema_version}.json"
    else:
        metadata_file = "data/metadata_schema_experiment_latest.json"
```

charles-turner-1: I knew that if chain looked iffy (haha), just wasn't seeing it for whatever reason 😅 .

Yeah, suspect it's not worth mucking with yet, although if we wind up in a situation with many more schemas (hopefully not) it might be worth rethinking.

joshuatorrance (approved):
Looks good! Thanks @charles-turner-1

PR #686, opened 2026-09-15
Add workflow to mirror catalog to Nectar

charles-turner-1:
## What

Adds a `workflow_dispatch` workflow (`.github/workflows/mirror_to_nectar.yml`) that mirrors the intake catalog from Gadi to Nectar object storage.

Unlike `gadi_deploy.yml`, this **does not** submit a PBS job via `qsub`/`appleboy`. Instead it runs `mirror-to-cloud` directly on the runner, which:
1. Pulls the catalog from Gadi over SSH (Fabric, `xp65_ci-dm` host alias) to the runner.
2. Uploads it from the runner to Nectar object storage.

## Inputs

- `catalog_version` (optional) — `vYYYY-MM-DD`; defaults to today.
- `hidden` (optional) — mirror as a hidden version (leading dot).

## Required secrets

- `DEPLOY_KEY` — private key authorised for the `xp65_ci` account on Gadi.
- `NECTAR_CLOUDS_YAML` — OpenStack `clouds.yaml` contents; the cloud must be named `nectar` (matches `openstack.connect(cloud="nectar")`).

## Notes

- SSH `known_hosts` is seeded via `ssh-keyscan` (trust-on-first-use). Can be pinned if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012LjKUo5H3Te1Nx1QjGHxai


charles-turner-1:
@joshuatorrance Unfortunately there's no way of triggering a manual CI run without merging the workflow to main first, so we might need a couple rounds of rubberstamp > test > rubberstamp to get this right. Assuming you don't spot any obvious issues.

joshuatorrance:
> @joshuatorrance Unfortunately there's no way of triggering a manual CI run without merging the workflow to main first, so we might need a couple rounds of rubberstamp > test > rubberstamp to get this right. Assuming you don't spot any obvious issues.

If you want to test it before merging you can temporarily add a `pull_request` trigger to trigger it from this branch right?

EDIT: oh you might need to mess about to get the `inputs` working with that method....

charles-turner-1:
Yeah, exactly - that works but not for a manual dispatch AFAIK

.github/workflows/mirror_to_nectar.yml
```diff
@@ -0,0 +1,79 @@
...
+            echo "::error::catalog_version must be in YYYY-MM-DD format (got '$CATALOG_VERSION')."
+            exit 1
+          fi
+          if ! date -d "$CATALOG_VERSION" >/dev/null 2>&1; then
+            echo "::error::catalog_version '$CATALOG_VERSION' is not a valid calendar date."
+            exit 1
+          fi
+
+      - name: Checkout code
+        uses: actions/checkout@v4
```
joshuatorrance: Any reason not to use the [most recent version of checkout](https://github.com/actions/checkout)?


```suggestion
        uses: actions/checkout@v7
```

.github/workflows/mirror_to_nectar.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Configure Nectar credentials
+        env:
+          NECTAR_CLOUDS_YAML: ${{ secrets.NECTAR_CLOUDS_YAML }}
+        run: |
+          mkdir -p ~/.config/openstack
+          printf '%s\n' "$NECTAR_CLOUDS_YAML" > ~/.config/openstack/clouds.yaml
+          chmod 600 ~/.config/openstack/clouds.yaml
+
+      - name: Setup pixi
+        uses: prefix-dev/setup-pixi@d3f436a425481402e6a95a1d1fc10331c708cd9e
```
joshuatorrance:
```suggestion
        uses: prefix-dev/setup-pixi@d3f436a425481402e6a95a1d1fc10331c708cd9e # v0.10.2
```

joshuatorrance (approved):
I don't see any obvious issues, assuming all the fiddly bash scripting works as intended.

Good luck!

charles-turner-1:
🤞

